PR #33, opened 2024-06-11
feat: add integration tests & integration test framework

obs-gh-nikhildua:
### Description

This PR adds integration tests and establishes an integration test framework using terraform tests and python, for observe-agent testing.

### Checklist


.github/workflows/tests-integration.yaml
```diff
@@ -0,0 +1,115 @@
...
+        
+        - name: Setup Python
+          uses: actions/setup-python@v5
+          with:
+            python-version: '3.10' 
+        - run: pip install -r scripts/requirements.txt           
+        
+        - name: Setup SSH Keys
+          run: |
+            echo "${{ secrets.SSH_PRIVATE_KEY }}" > ${{env.TF_VAR_PRIVATE_KEY_PATH}}
```
obs-gh-alexlew, line 83: what are these needed for?

obs-gh-alexlew: for AWS access?

integration/modules/create_ec2/main.tf
```diff
@@ -0,0 +1,57 @@
...
+    local.BASE_TAGS,
+    {
+      Name = format(var.name_format, "publicKey_${var.AWS_MACHINE}_${random_string.output.id}")
+    },
+  )
+
+}
+
+#Create EC2 instance for Observe Agent Testing
+resource "aws_instance" "observe_agent_instance" {
```
obs-gh-alexlew, line 25: can we tag these with some attribute indicating that they're for agent integration tests?

integration/modules/create_ec2/outputs.tf
```diff
@@ -0,0 +1,42 @@
...
+output "instance_state" {
+  value = aws_instance.observe_agent_instance.instance_state
+}
+
+
+output "public_ip" {
+  value = aws_instance.observe_agent_instance.public_ip
+}
+
+
```
obs-gh-alexlew, line 34: nit: some inconsistent whitespace, 2 newlines

integration/modules/create_ec2/user_data/yumbased.sh
```diff
@@ -0,0 +1,18 @@
...
+
+yum install curl -y
+
+yum install wget -y
+
+yum install ca-certificates -y
+
+
+
+# echo '[fury]
```
obs-gh-alexlew, line 12: we can rm this for now, its stored in the instructions so no worries about losing it

integration/modules/setup_aws_backend/README.md
```diff
@@ -0,0 +1,20 @@
+## setup_aws_backend module
+
+The module performs the following actions in `us-west-1` region. It's meant to be only used once as an initial setup for state management.
```
obs-gh-alexlew, line 3: only once ever? or only once per run

integration/modules/setup_aws_backend/provider.tf
```diff
@@ -0,0 +1,12 @@
+provider "aws" {
+  region  = "us-west-1"
+  profile = "blunderdome"
+  assume_role {
+    role_arn = "arn:aws:iam::767397788203:role/OrganizationAccountAccessRole"
+  }
+}
+
+
+# provider "aws" {
```
obs-gh-alexlew, line 10: rm

integration/modules/setup_aws_backend/role.tf
```diff
@@ -0,0 +1,131 @@
...
+      test     = "StringLike"
+      variable = "token.actions.githubusercontent.com:sub"
+      values = [
+        # lock to repo, but match any branch
+        "repo:observeinc/observe-agent:*",
+      ]
+    }
+  }
+
+  #Nikhil-PS Account (767397788203)
```
obs-gh-alexlew, line 54: whats this role?

integration/modules/setup_aws_net_sg/README.md
```diff
@@ -0,0 +1,24 @@
+## setup_aws_net_sg module
+
+This module sets up the AWS networking and security groups for the integration tests. It's meant to be only setup once for integration tests to use when creating and destroying EC2 instances
```
obs-gh-alexlew, line 3: if this is just meant to be run once, why is it a separate module from the other setup module?

integration/modules/create_ec2/locals.tf
```diff
@@ -0,0 +1,118 @@
...
+}
+
+
+#Map of tags we'll attach
+locals {
+  BASE_TAGS = {
+    owner        = "Observe"
+    createdBy    = "terraform"
+    team         = "Product Specialists "
+    purpose      = "observe-agent integration tests"
```
obs-gh-alexlew, line 115: nice!

integration/modules/create_ec2/main.tf
```diff
@@ -0,0 +1,57 @@
+
+# Random String for naming
+resource "random_string" "output" {
```
obs-gh-alexlew, line 3: maybe call this `unique_identifier` or something

integration/scripts/test_diagnosis.py
```diff
@@ -0,0 +1,75 @@
...
+    init_command='sudo observe-agent init-config --token {} --observe_url {}'.format(env_vars["observe_token"], env_vars["observe_url"])
+    diagnose_command='observe-agent diagnose'
+    config_file_linux = '/etc/observe-agent/observe-agent.yaml'
+
+    #Set up correct config with observe url and token 
+    result = remote_host.run_command(init_command)
+    write = False
+    use = False
+    for line in result.stdout.splitlines():      
+        if "Writing configuration values to {}".format(config_file_linux) in line:
```
obs-gh-alexlew, line 33: honestly i think this might be a little brittle, if we ever change the string output from this command we'll need to change this test too. checking for the exact string seems like something a unit test would do, here we just want to see if things are working more generally. maybe we can relax this/other checks like this? open to thoughts

integration/scripts/test_status.py
```diff
@@ -0,0 +1,66 @@
...
+    """ 
+
+   start_command='sudo systemctl enable --now observe-agent'
+   status_command='observe-agent status'
+   metrics_dict = defaultdict(list)
+
+
+   #Start Observe Agent 
+   remote_host.run_command(start_command)
+   time.sleep(5)
```
obs-gh-alexlew, line 30: maybe instead of a hardcoded sleep we can do a while loop with a limit on the number of runs

integration/scripts/test_version.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
+import os
+import sys
+import re
+import time 
+from utils import *
+
+
+@print_test_decorator
+def run_test_linux(remote_host: Host, env_vars: dict) -> None:
```
obs-gh-alexlew, line 10: do these tests generally assume that the agent is already installed on the host?

obs-gh-alexlew:
are we cleaning up the ec2 instances afterwards? can you point me at that part?

.github/workflows/tests-integration.yaml
obs-gh-nikhildua: Yeap. This maps the secret and converts it a respective `.pem` file for EC2 access

integration/modules/setup_aws_backend/README.md
obs-gh-nikhildua: The `setup_aws_backend` and `setup_aws_net_sg` are only ever used once to setup the correct "playground" for ec2 instances that will be created

Technically they are not part of integration test flow but I didn't know where else to put the terraform code for future reference

integration/modules/setup_aws_backend/role.tf
obs-gh-nikhildua: Updating the comment:
"This is created in member account Nikhil-PS Account (767397788203) in the blunderdome root account for billing"

integration/modules/setup_aws_net_sg/README.md
obs-gh-nikhildua: See https://github.com/observeinc/observe-agent/pull/33#discussion_r1655534883

Only `create_ec2` is used in our tests. See https://github.com/observeinc/observe-agent/blob/nikhil/integration-tests-poc/integration/tests/integration.tftest.hcl#L4

obs-gh-nikhildua:
> are we cleaning up the ec2 instances afterwards? can you point me at that part?

Yeap. Tf tests automatically sets up and destroys as part of in-memory state 

https://developer.hashicorp.com/terraform/tutorials/configuration-language/test

"Terraform tests let you validate your module configuration without impacting your existing state file or resources. Testing is a separate operation that is not part of a plan or apply workflow, but instead builds ephemeral infrastructure and tests your assertions against in-memory state for those short-lived resources"

integration/scripts/test_version.py
obs-gh-nikhildua: Yes the tests follow this flow 

Installation happens first and we check that's valid before moving onto next test for version 

https://github.com/observeinc/observe-agent/blob/nikhil/integration-tests-poc/integration/tests/integration.tftest.hcl

integration/scripts/test_diagnosis.py
obs-gh-nikhildua: Would you suggest just looking to make sure it wrote some`observe-agent.yaml` ?

integration/scripts/test_status.py
obs-gh-nikhildua: Added

integration/modules/setup_aws_backend/role.tf
obs-gh-alexlew: is there some sort of service account we can use for this?

obs-gh-alexlew (approved):

integration/modules/setup_aws_backend/role.tf
obs-gh-nikhildua: Yeap. We are actually using a "service account"! It's just an IAM role in this member account

PR #51, opened 2024-07-15
feat: add docker image tests

obs-gh-nikhildua:
### Description

Ticket: https://observe.atlassian.net/browse/OB-33428

- Adds docker image integration tests 
- Updates to support `cloud-init` for Ubuntu (debian) machine for docker purposes 
- Docker images are also now part of build step, for upload. They are packaged as `.tar` for each distribution to "install" on the debian machine 
(See https://docs.docker.com/reference/cli/docker/image/save/ & https://docs.docker.com/reference/cli/docker/image/load/) 
- Minor wrapper helper function for common work between windows, linux, docker 

### Checklist


integration/scripts/utils.py
```diff
@@ -10,6 +10,52 @@
 import json 
 import pprint 
 
+
+
+def create_default_config_file(destination_file_path: str = "/tmp/observe-agent.yaml") -> None:
+    
+    """
+    Creates a defailt observe-agent.yaml at destination. Useful for docker testing since this is not mounted
```
obs-gh-alexlew, line 18: nit typo on default

integration/scripts/utils.py
```diff
@@ -10,6 +10,52 @@
...
+    
+    """
+    Creates a defailt observe-agent.yaml at destination. Useful for docker testing since this is not mounted 
+
+    Args:
+       destination_file_path (str, optional): path to create the config file. Defaults to "/tmp/observe-agent.yaml".
+
+    """
+
+    file_content = """
```
obs-gh-alexlew, line 25: any way we can copy this out of an image package vs. hard coding it here? as we change the defaults itd be nice to not have to change it here as well

obs-gh-alexlew: that being said we already have the defaults saved in 3 different places so if its just one more place its not a big deal, this ones just easier to miss/its not obvious this needs to be updated as well

obs-gh-nikhildua: Currently the file is not part of https://github.com/observeinc/observe-agent/tree/main/packaging/docker/observe-agent correct? 
Would this have to be added in packaging somewhere?

obs-gh-nikhildua: Updated

obs-gh-alexlew (approved):

PR #74, opened 2024-08-15
OB-34960 Decorate k8s Node facets with custom processors

obs-gh-enricogiorio:
Add "Status" and "Roles" facets

Rework the testing a bit to allow using jmespath to query specific fields of interest instead of manually creating json-like objects in go to exactly match the output.

### Description

This PR adds the "Roles" and "Status" facets to Node.
Since "Roles" is an array, I had to change the interface of the existing `ValueFn()` to return `any` instead of `string`, so that we can return maps and arrays, too.

The current testing process is a bit cumbersome and doesn't scale well, since we are making a single `pcommon.Logs` containing all testing input logs, and then we have to maintain a parallel slice of "desired outputs", correlating inLogs[i] with outResults[i]. Since the correlation is always 1:1, I figured I could change this to:
each testcase contains a single `pcommon.Logs`, which contains only ONE Log. And each testcase also contains whatever verification(s) we want for the processed output of such log. Then we simply iterate over the testcases.
I also got rid of all manually-created `map[string]interface{}` json-like madness and used `jmespath`, which allows for simpler and more powerful queries, so that we can selectively test what we need from the processed Logs!
Given that we don't do correlation between logs, I figured that the assumption of always having a single log both as input and as output of the processor is enough for what we are doing here.

### Checklist
- [ ] Created tests which fail without the change (if possible) --> not possible, this is modular

### DIFF
--- a/components/processors/observek8sattributesprocessor/processor.go
+++ b/components/processors/observek8sattributesprocessor/processor.go
@@ -3,6 +3,7 @@ package observek8sattributesprocessor
 import (
 	"context"
 	"encoding/json"
+	"errors"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/pdata/pcommon"
@@ -21,10 +22,13 @@ type K8sEventsProcessor struct {
 	actions []K8sEventProcessorAction
 }
 
+type attributes map[string]any
+
+type transform func(plog.LogRecord) (attributes, error)
+
 type K8sEventProcessorAction struct {
-	Key      string
-	ValueFn  func(plog.LogRecord) any
-	FilterFn func(K8sEvent) bool
+	ComputeAttributes transform
+	FilterFn          func(K8sEvent) bool
 }
 
 func newK8sEventsProcessor(logger *zap.Logger, cfg component.Config) *K8sEventsProcessor {
@@ -32,7 +36,7 @@ func newK8sEventsProcessor(logger *zap.Logger, cfg component.Config) *K8sEventsP
 		cfg:    cfg,
 		logger: logger,
 		actions: []K8sEventProcessorAction{
-			PodStatusAction, NodeStatusAction, NodeRolesAction,
+			PodStatusAction, NodeStatusAction, NodeRolesAction, PodContainersCountsAction,
 		},
 	}
 }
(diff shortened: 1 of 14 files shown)


obs-gh-enricogiorio:
I pushed a 2nd commit because I didn't know it would fall into the same PR, I thought it was gonna work the same as gerrit. If that's ok for you, we can leave it as-is and merge both together (it kinda makes sense as the second commit depends on the refactorings I made in the first one).

components/processors/observek8sattributesprocessor/processor.go
```diff
@@ -20,18 +22,21 @@ type K8sEventsProcessor struct {
...
+type transformFn func(plog.LogRecord) any
+
 type K8sEventProcessorAction struct {
-	Key      string
-	ValueFn  func(plog.LogRecord) string
-	FilterFn func(K8sEvent) bool
+	Key                    string
+	Transform              transformFn
+	FilterFn               func(K8sEvent) bool
+	ComputesMultipleFacets bool
```
obs-gh-alexlew, line 31: I think instead of branching the logic on this field we should just refactor the existing single-facet actions to all just return a map with only one key-value pair and have the action's `Transform` (maybe a better name for the fn would be something like `ComputeAttributes`) function just always return a map. And we'd drop the `Key` field.

obs-gh-enricogiorio: I did it like that before, but it doesn't work because what if we want to compute a json object as attribute? How can we distinguish if the map is a list of facets or a facet that is a JSON object?
What you said is exactly what I did before the force push above your message, then I realised that it might be too much of a limitation (not being able to generate JSON object facets), so I made this new "branching".

obs-gh-enricogiorio: I will rename the interface to `ComputeAttributes` in the next PR (it's already made, I just need to push it). YOu're right, it's a better naming

obs-gh-alexlew: The attributes returned would be of type `map[string]any` so the value we're setting at a given attribute key could be either a string or a json object. not sure how this change results in that limitation

components/processors/observek8sattributesprocessor/processor.go
```diff
@@ -55,32 +60,113 @@ func (kep *K8sEventsProcessor) processLogs(_ context.Context, logs plog.Logs) (p
...
+					// This is where the custom processor actually computes the transformed value(s)
+					values, err := action.ComputeAttributes(lr)
+					if err != nil {
+						return logs, err
+					}
+
+					facetsMap.EnsureCapacity(facetsMap.Len() + len(values))
+					for key, val := range values {
+						if err := mapPut(facetsMap, key, val); err != nil {
+							return logs, err
```
obs-gh-alexlew, line 100: should we give up when we hit an error and return the logs in the current state? or should we just skip that action and continue

obs-gh-enricogiorio: Aha good catch. Let's continue instead of aborting the whole thing!

obs-gh-alexlew (approved):

PR #83, opened 2024-08-23
OB-35830 Add Node "nodePool" facet

obs-gh-enricogiorio:
This facet is present only when the kubernetes deployment is a managed service by one of the following major providers:

- AWS EKS
- GCP EKE
- Azure AKS
- DigitalOcean DOKS

### DIFF
--- a/components/processors/observek8sattributesprocessor/testdata/nodeObjectEventSimple.json
+++ b/components/processors/observek8sattributesprocessor/testdata/nodeObjectEventSimple.json
@@ -22,7 +22,8 @@
             "minikube.k8s.io/updated_at": "2024_08_13T16_47_02_0700",
             "minikube.k8s.io/version": "v1.33.1",
             "node-role.kubernetes.io/control-plane": "",
-            "node.kubernetes.io/exclude-from-external-load-balancers": ""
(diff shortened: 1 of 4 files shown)


components/processors/observek8sattributesprocessor/testdata/nodeObjectEventSimple.json
```diff
@@ -22,7 +22,8 @@
             "minikube.k8s.io/updated_at": "2024_08_13T16_47_02_0700",
             "minikube.k8s.io/version": "v1.33.1",
             "node-role.kubernetes.io/control-plane": "",
-            "node.kubernetes.io/exclude-from-external-load-balancers": ""
+            "node.kubernetes.io/exclude-from-external-load-balancers": "",
+            "eks.amazonaws.com/nodegroup": "test-node-group"
```
obs-gh-enricogiorio, line 26: This entry is taken from an event ingested in O2 from our very own EKS cluster. I only changed the value of the label to prevent sensitive information leak.
The key is not copied from the agent code I wrote to confirm that the EKS key I put in the code is indeed correct.

xueweiz (approved):

PR #84, opened 2024-08-28
feat: add custom actions for Workload entities

obs-gh-enricogiorio:
Consolidate actions into fewer files: one file per entity.

Get rid of the common interface for actions as it serves no real purpose.

Type-cast objects only once after unmarshalling (rather than once per action).

Consolidate actions' interfaces to accept the right type for such action:

 - PodAction accepts v1.Pod
 - NodeAction accepts v1.Node,
 - ... actions for new entity types will do the same

### Description

### Checklist


components/processors/observek8sattributesprocessor/nodeactions_test.go
```diff
@@ -0,0 +1,48 @@
+package observek8sattributesprocessor
+
+import "testing"
+
+func TestNodeActions(t *testing.T) {
+	for _, testCase := range []k8sEventProcessorTest{
+		{
+			name: "Node Status and Role",
```
obs-gh-alexlew, line 8: kind of interesting that the test cases span across the actual actions, wouldn't it make more sense for them to be distinct? so each test case only applies to one action

obs-gh-alexlew: is it just to reduce duplication of test events?

components/processors/observek8sattributesprocessor/podactions.go
```diff
@@ -0,0 +1,191 @@
...
+	PodTotalContainersAttributeKey   = "total_containers"
+	PodReadyContainersAttributeKey   = "ready_containers"
+
+	PodReadinessGatesReadyAttributeKey = "readinessGatesReady"
+	PodReadinessGatesTotalAttributeKey = "readinessGatesTotal"
+
+	PodConditionsAttributeKey = "conditions"
+)
+
+type PodStatusAction struct{}
```
obs-gh-alexlew, line 24: nit: seems like this is missing the same header comment that other actions have

components/processors/observek8sattributesprocessor/types.go
```diff
@@ -23,22 +27,47 @@ type K8sEventProcessorAction interface {
...
+	for _, action := range m.podActions {
+		atts, err := action.ComputeAttributes(pod)
+		if err != nil {
+			return res, err
+		}
+		res.addAttributes(atts)
 	}
-	return node, nil
+	return res, nil
 }
```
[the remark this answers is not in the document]
obs-gh-alexlew: i like this better!

components/processors/observek8sattributesprocessor/nodeactions_test.go
obs-gh-enricogiorio: Mmm yeah I didn't really create these, they were already present in processor_test from previous PRs.
Maybe it's better to divide them in another commit, it's not really the purpose of this PR. I just moved them around here.

components/processors/observek8sattributesprocessor/podactions.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	}
+
+	return attributes{
+		PodReadinessGatesTotalAttributeKey: len(pod.Spec.ReadinessGates),
+		PodReadinessGatesReadyAttributeKey: readinessGatesReady,
+	}, nil
+}
+
+// ---------------------------------- Pod "readiness" ----------------------------------
```
obs-gh-alexlew, line 171: nit, this should be Pod "conditions" it looks like its a dupe of the header above

obs-gh-alexlew (approved):

PR #88, opened 2024-09-06
feat: implement custom processor actions for all facets that involve calculating the length of a pcommon.Slice

obs-gh-enricogiorio:
OTTL's `Len()` doesn't work well with some of the slices obtained by unmarshalling k8s events into their respective API entities. This is because `Len()` doesn't support the type of the inner elements of most of the slices we have to use.
As a result, the only way to "calculate" such lengths is via custom code in our events' processor.

### DIFF
--- a/components/processors/observek8sattributesprocessor/podactions_test.go
+++ b/components/processors/observek8sattributesprocessor/podactions_test.go
@@ -79,6 +79,18 @@ func TestPodActions(t *testing.T) {
 				{"observe_transform.facets.conditions.TestCondition", "Unknown"},
 			},
 		},
+		{
+			name: "Pod conditions",
+			inLogs: createResourceLogs(
+				logWithResource{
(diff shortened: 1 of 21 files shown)


components/processors/observek8sattributesprocessor/podactions_test.go
```diff
@@ -5,78 +5,90 @@ import "testing"
...
-				},
-			),
-			expectedResults: []queryWithResult{
-				{"observe_transform.facets.readinessGatesReady", int64(1)},
-				{"observe_transform.facets.readinessGatesTotal", int64(2)},
-			},
-		},
+		// {
+		// 	name: "noObserveTransformAttributes",
+		// 	inLogs: createResourceLogs(
```
xueweiz, line 10: I guess these are not intentional?

xueweiz (approved):

PR #91, opened 2024-09-10
fix: update file copying to work for upgrades

obs-gh-alexlew:
### Description

OB-36354 Modify windows install script copying files to work the same for initial install and upgrading

### Checklist


obs-gh-alexlew:
old code was handling when folder exists and when it doesn't differently, this should match the behavior to just fully overwrite

arthur-observe (approved):

PR #94, opened 2024-09-23
build(deps): add resourceprocessor

obs-gh-alexlew:
### Description

OB-36942 Add resource processor

### Checklist


xueweiz (approved):
Aha I see now how it's done. Basically:
1. add dependency in go.mod
2. go mod vendor stuff
3. register it in components.go explicitly

Curious, how did you generate the README.md update? (If there is an instruction documented somewhere, a link would be perfect)

(Next time if I need something, I can try not to bother you writing it out and just ask you to review it

obs-gh-alexlew:
There's instructions on updating the component here: https://github.com/observeinc/observe-agent?tab=readme-ov-file#adding-new-components 

its actually a little complicated since we need to add it to the manifest and then install the `ocb` tool and use that in the makefile command to generate the new files and move them to the right places. The README.md i'm manually updating by hand using https://www.tablesgenerator.com/markdown_tables

PR #102, opened 2024-10-04
fix: OB-37389 Obfuscate secrets inside Secret previous configuration

obs-gh-enricogiorio:
Secret entities not only contain secrets in data, but might also contain
an annotation that represents the previous configuration for this
Secret, including the secrets. Drop such annotation

### DIFF
--- a/components/processors/observek8sattributesprocessor/secretactions.go
+++ b/components/processors/observek8sattributesprocessor/secretactions.go
@@ -16,10 +16,21 @@ func NewSecretRedactorBodyAction() SecretRedactorBodyAction {
 
 // ---------------------------------- Secret "data" values' redaction ----------------------------------
 
-// Redacts secrets' values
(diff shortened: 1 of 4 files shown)


components/processors/observek8sattributesprocessor/secretactions.go
```diff
@@ -8,6 +8,16 @@ const (
 	RedactedSecretValue = "REDACTED"
 )
 
+var keysMightContainSecret map[string]struct{}
+
+func init() {
```
xueweiz, line 13: I think you don't need these anymore?

xueweiz (approved):

components/processors/observek8sattributesprocessor/secretactions.go
obs-gh-enricogiorio: Good catch

xueweiz (approved):

PR #106, opened 2024-10-17
feat: refactor diagnose command to be more extensible

obs-gh-mattcotter:
### Description

OB-37709 Refactor in preparation to add config yaml diagnosis.

### Checklist


internal/commands/diagnose/diagnose.go
```diff
@@ -22,38 +22,48 @@ var (
...
+		for _, diagnostic := range diagnostics {
+			data, err := diagnostic.check()
+			if err != nil {
+				return err
+			}
+			t := template.Must(template.
+				New(diagnostic.templateName).
+				ParseFS(diagnostic.templateFS, diagnostic.templateName))
+			if err := t.ExecuteTemplate(os.Stdout, diagnostic.templateName, data); err != nil {
+				return err
```
obs-gh-alexlew, line 54: if a specific check fails, we might not want to return the error and skip the rest of the checks. maybe it would be better to have some sort of default error message that we print out

internal/commands/diagnose/diagnose.go
```diff
@@ -22,38 +22,48 @@ var (
...
-	// and all subcommands, e.g.:
-	// diagnoseCmd.PersistentFlags().String("foo", "", "A help for foo")
-
-	// Cobra supports local flags which will only run when this command
-	// is called directly, e.g.:
-	// diagnoseCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
 }
 
-func runNetworkCheck() error {
+func authCheck() (any, error) {
```
obs-gh-alexlew, line 65: can we move this to a different file? I'm imagining one file per check

obs-gh-mattcotter: Updated! Here's how it looks like to add the yaml parse check now:
https://github.com/observeinc/observe-agent/pull/107/files#diff-ab787288156b418b841c779f91a068907aade153aec12c3cd8ab82c841c00fc6R25

internal/commands/diagnose/diagnose.go
obs-gh-mattcotter: Good idea! I added warning icons so it's more clear when there's a failure, but lmk if we're trying to stick to ASCII and I can remove those

internal/commands/diagnose/networkchecks.go
obs-gh-alexlew, line 1: maybe we can just rename this authcheck.go since that's really what it is and we might be adding some separate networking checks later

obs-gh-alexlew (approved):

PR #107, opened 2024-10-17
feat: add YAML parse check to diagnose command

obs-gh-mattcotter:
### Description

OB-37709 add a new check to the diagnose command to check if the config file is valid YAML.

### Checklist

### DIFF
--- a/internal/commands/diagnose/configcheck.go
+++ b/internal/commands/diagnose/configcheck.go
@@ -16,7 +16,7 @@ type ConfigTestResult struct {
 }
 
 func validateYaml(yamlContent []byte) error {
-	m := make(map[any]any)
+	m := make(map[string]any)
 	return yaml.Unmarshal(yamlContent, &m)
 }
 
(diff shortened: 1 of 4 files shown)


internal/commands/diagnose/configcheck.go
```diff
@@ -0,0 +1,58 @@
...
+	Error      string
+}
+
+func validateYaml(yamlContent []byte) error {
+	m := make(map[string]any)
+	return yaml.Unmarshal(yamlContent, &m)
+}
+
+func checkConfig() (any, error) {
+	if root.CfgFile == "" {
```
obs-gh-alexlew, line 24: where does this get set? would `viper.ConfigFileUsed()` be a better fit here?

obs-gh-mattcotter: This gets set by Cobra here:
https://github.com/observeinc/observe-agent/blob/9b75d11e06f8801ee7b41c24cc4573db1b2fef7d/internal/root/root.go#L37

But you're right, we have custom handling to find the default file in Viper if that's not set. I updated to `viper.ConfigFileUsed()`, good catch!

obs-gh-alexlew (approved):

PR #113, opened 2024-10-23
feat: add new command to send test data to Observe

obs-gh-mattcotter:
### Description

OB-32242 Create a new sub command to send test JSON data directly to Observe via HTTP post API.

Example usage:
```bash
$ ./observe-agent send-test-data --data '{"testKey": "test data"}'
Using config file: /etc/observe-agent/observe-agent.yaml
Successfully sent test data. Saw response: {"ok":true}
```
Example result:
Screenshot 2024-10-23 at 2 10 24 PM

### Checklist


internal/commands/sendtestdata/sendtestdata.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/observeinc/observe-agent/internal/root"
+	"github.com/spf13/cobra"
+)
+
+const TestDataPath = "/observe-agent/test"
+
+var defaultTestData = map[string]any{
+	"hello": "world",
+}
```
obs-gh-mattcotter, line 18: Are these defaults okay?

internal/commands/sendtestdata/postdata.go
```diff
@@ -0,0 +1,48 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/spf13/viper"
+)
+
+func PostTestData(data any, URL string, headers map[string]string) (string, error) {
```
obs-gh-mattcotter, line 14: I would happily take suggestions of how to unit test this!

We could use https://github.com/jarcoal/httpmock, but I'm not sure it's worth adding a dependency. Thoughts?

internal/commands/sendtestdata/sendtestdata.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/observeinc/observe-agent/internal/root"
+	"github.com/spf13/cobra"
+)
+
+const TestDataPath = "/observe-agent/test"
```
obs-gh-alexlew, line 14:
```
curl https://175914298205.collect.observe-eng.com/v1/http \
 -H "Authorization: Bearer $TOKEN" \
 -H "Content-type: application/json" \
 -d '{"data": "hello world"}'
 ```

request should look like this

internal/commands/sendtestdata/postdata.go
obs-gh-alexlew: think its fine to add a dep for this test!

internal/commands/sendtestdata/sendtestdata.go
obs-gh-mattcotter: That is [how the POST request is structured](https://github.com/observeinc/observe-agent/pull/113/files#diff-b52112d16afa72bb2394fce39570897bc3971814dc00338ea9ef3592ba4b7e06R44-R46), but I was getting a `path` item of `"/"` in `extra` even when not specifying a path, so I added this for clarity. You can see my example screenshot in the PR description. Lmk if there's a way to exclude the `extra` all together! I couldn't find it in [our docs](https://docs.observeinc.com/en/latest/content/data-ingestion/endpoints/http.html#ingesting-data-over-http) though.

obs-gh-mattcotter: Hopefully the test I added can clarify this! If not, I might be misunderstanding your request.

obs-gh-enricogiorio: I don't understand why you are sending to https://observe-agent/test/ , I think maybe Alex meant that you can send the testdata to a hardcoded collect path of that specific test customer?

internal/commands/sendtestdata/sendtestdata.go
obs-gh-enricogiorio: Yes totally ok

obs-gh-enricogiorio (commented):
LGTM, just double check the hardcoded URL

internal/commands/sendtestdata/sendtestdata.go
obs-gh-mattcotter: I renamed this to extra path; the post is being made to eg. `https://175914298205.collect.observe-eng.com/v1/http/observe-agent/test`. In our http API, it seems that you always get a field in the `extra` data with the path that comes after `/v1/http`, so I wanted to have something that stuck out and made it clear instead of just `/`. Does that clarify this?

obs-gh-enricogiorio (approved):

PR #115, opened 2024-10-24
feat: validate observe-agent config instead of simply parsing the yaml

obs-gh-mattcotter:
### Description

OB-31850 Add validation of observe-agent config on top of simple YAML parsing.

### Checklist


internal/commands/diagnose/configcheck.tmpl
```diff
@@ -1,6 +1,6 @@
-Running check on config file {{ .ConfigFile }}
+Running check on observe-agent config file {{ .ConfigFile }}
 {{- if .Passed }}
-Config file is valid YAML.
```
obs-gh-alexlew, line 3: feels like if it errors for invalid yaml vs. erroring for failing validation those should be reflected separately in the template

internal/config/configschema.go
```diff
@@ -0,0 +1,67 @@
+package config
+
+import (
+	"errors"
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+type HostMonitoringLogsConfig struct {
```
obs-gh-alexlew, line 10: a lot of these I defined already in https://github.com/observeinc/observe-agent/blob/main/internal/commands/initconfig/configschema.go i like it being in config/configschema.go can you remove the definitions I have then and rework those references?

obs-gh-mattcotter: I did that! `internal/commands/initconfig/configschema.go` is removed in this diff (not sure why git didn't register it as a move).

obs-gh-alexlew (approved):

PR #121, opened 2024-10-31
refactor: clean up dead testing code

obs-gh-enricogiorio:
Improve the testing logic to allow each testcase to specify which part of the logRecord the jmespath query should be executed, rather than deciding it once for all testcases of the test function.

Remove unused fields + logic that allowed testing w/ custom functions, since we can do everything with jmespath


components/processors/observek8sattributesprocessor/processor.go
```diff
@@ -316,13 +316,16 @@ func (kep *K8sEventsProcessor) processLogs(_ context.Context, logs plog.Logs) (p
...
 				}
+
 				facets, exists := transformMap.Get("facets")
-				if exists {
-					facetsMap = facets.Map()
-				} else {
-					facetsMap = transformMap.PutEmptyMap("facets")
+				if !exists {
+					kep.logger.Error("no facets map in observe_transform", zap.Error(err))
+					continue
```
obs-gh-mattcotter, line 326: This is a functional change and not just a refactor, right? Now if there is an `observe_transform` attribute with no `facets`, it will error and ignore rather than defaulting to empty. Is this what we want?

obs-gh-mattcotter (approved):

components/processors/observek8sattributesprocessor/processor.go
obs-gh-enricogiorio: Actually I made a mistake here, I should have left it as it was. This is dangerous, will revert this bit.

obs-gh-enricogiorio: Because I thought that observe_transform could only be created here (and so we can assume that facets will be here as well).
But actually, this could be created in helm as well.
But the catch here is that if observe_transform is present, then facets is also present inside it

PR #130, opened 2024-11-13
fix: skip deb and rpm publishing on nightly release

obs-gh-mattcotter:
### Description

OB-38483 skip deb and rpm publishing on nightly release

I will test this by running the action when this change lands.

### Checklist


obs-gh-alexlew (approved):

obs-gh-enricogiorio:
I didn't even know nightly builds were available via `apt upgrade`!
I also think they shouldn't be, so I'm totally on board with this approach.

PR #132, opened 2024-11-14
fix: allow OTEL_CONFIG_OVERRIDES env var to replace agent config overrides

obs-gh-mattcotter:
### Description

OB-38640 allow OTEL_CONFIG_OVERRIDES env var to replace agent config overrides

This will allow the agent to be fully run with env variables only.

example:
```sh
$ OTEL_CONFIG_OVERRIDES=`cat tmp.yml` ./observe-agent config
# ======== computed agent config
host_monitoring:
    enabled: true
    logs:
        enabled: true
        include: []
    metrics:
        host:
            enabled: true
        process:
            enabled: false
observe_url: ""
otelconfigfile: ""
self_monitoring:
    enabled: true
token: ""


# ======== config file /var/folders/y8/v7cyk9611qbg059c0q02vk0w0000gn/T/observe-agent2529970453/otel-config-overrides-4110568018.yaml
processors:
    probabilistic_sampler:
        sampling_percentage: 50
service:
    pipelines:
        logs/forward:
            exporters:
                - debug
            processors:
                - probabilistic_sampler
            receivers:
                - otlp
```

### Checklist

### DIFF
--- a/internal/connections/confighandler.go
+++ b/internal/connections/confighandler.go
@@ -11,30 +11,34 @@ import (
 	logger "github.com/observeinc/observe-agent/internal/commands/util"
 	"github.com/observeinc/observe-agent/internal/config"
 	"github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
 )
 
-func GetAllOtelConfigFilePaths(ctx context.Context, tmpDir string) ([]string, string, error) {
+const (
+	OTEL_OVERRIDE_YAML_KEY = "otel_config_overrides"
+)
+
+func GetAllOtelConfigFilePaths(ctx context.Context, tmpDir string) ([]string, error) {
 	configFilePaths := []string{}
 	// If the default otel-collector.yaml exists, add it to the list of config files
 	defaultOtelConfigPath := filepath.Join(GetDefaultConfigFolder(), "otel-collector.yaml")
 	if _, err := os.Stat(defaultOtelConfigPath); err == nil {
 		agentConf, err := config.AgentConfigFromViper(viper.GetViper())
 		if err != nil {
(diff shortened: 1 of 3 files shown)


internal/connections/confighandler.go
```diff
@@ -44,16 +48,32 @@ func GetAllOtelConfigFilePaths(ctx context.Context, tmpDir string) ([]string, st
...
+	if viper.IsSet(OTEL_OVERRIDE_YAML_KEY) {
+		// GetStringMap is more lenient with respect to conversions than Sub, which only handles maps.
+		overrides := viper.GetStringMap(OTEL_OVERRIDE_YAML_KEY)
+		if len(overrides) == 0 {
+			stringData := viper.GetString(OTEL_OVERRIDE_YAML_KEY)
+			// If this was truly set to empty, then ignore it.
+			if stringData != "" {
+				// Viper can handle overrides set in the agent config, or passed in as an env var as a JSON string.
+				// For consistency, we also want to accept an env var as a YAML string.
+				err := yaml.Unmarshal([]byte(stringData), &overrides)
```
obs-gh-alexlew, line 60: does this handle nested configs correctly? we have a different delimiter `::` for nested keys

obs-gh-alexlew: https://opentelemetry.io/docs/collector/configuration/#override-settings similar to how otel does it (this is because there can be periods within a key name in the config so that overloads '.' as the delimiter

obs-gh-alexlew:
so what's the actual value we provide in this env var? what does `cat tmp.yml` look like when expanded?

obs-gh-alexlew (approved):

PR #133, opened 2024-11-20
Makefile: install-ocb target -> add ~/bin directory creation before curl output…

obs-gh-scottsolmonson:
… targets it

### Description

Problem: "install-ocb" makefile target assumes existence of ~/bin directory - fails if doesn't exist
Fix: add non-destructive creation of ~/bin directory before curl tries to output there

### Checklist
- [NA] Created tests which fail without the change (if possible)
- [NA] Extended the README / documentation, if necessary

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,7 @@ release:
 
 ## install-ocb: Installs correct version of ocb binary
 install-ocb:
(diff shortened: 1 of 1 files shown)


obs-gh-scottsolmonson:
Closing due to incorrect workflow

PR #140, opened 2024-12-06
feat: add env var example to docker-compose, add agent config to image

obs-gh-mattcotter:
### Description

Add env var example to docker-compose, add agent config to image.
The goal of this change is to make Docker usage more clear.

### Checklist

### DIFF
--- a/packaging/docker/observe-agent/observe-agent.yaml
+++ b/packaging/docker/observe-agent/observe-agent.yaml
@@ -1,0 +1,13 @@
+# A minimal default agent config which should be overwritten via a docker volume or environment variables.
+debug: false
+self_monitoring:
+  enabled: true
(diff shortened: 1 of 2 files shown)


packaging/docker/observe-agent/observe-agent.yaml
```diff
@@ -0,0 +1,13 @@
+# A minimal default agent config which should be overwritten via a docker volume or environment variables.
```
obs-gh-mattcotter, line 1: Having this present in the image will avoid "file not found" issues when no alternative is provided via volume mount (ie when someone is running the image with env config only). If a volume is provided, that will be used instead of this.

obs-gh-alexlew (approved):

PR #141, opened 2024-12-09
feat: update mac install script to download from latest release

obs-gh-mattcotter:
### Description

Update mac install script to download from latest release. Now that we have a new release which builds the expected Mac zip, we can finally download and install from GitHub.

### Checklist

### DIFF
--- a/scripts/install_mac.sh
+++ b/scripts/install_mac.sh
@@ -6,6 +6,38 @@ service_name="com.observeinc.agent"
 observeagent_install_dir="/usr/local/observe-agent"
 tmp_dir="/tmp/observe-agent"
 
+# Parse args
+while [ $# -gt 0 ]; do
+    opt=$1
+    shift
+    arg=""
+    if [[ "$opt" == *"="* ]]; then
+        arg=$(echo $opt | cut -d'=' -f2)
+        opt=$(echo $opt | cut -d'=' -f1)
+    elif [ $# -gt 0 ]; then
+        arg="$1"
+        shift
+    fi
+    case "$opt" in
+        --token)
+            TOKEN="$arg"
+            ;;
+        --observe_url)
+            OBSERVE_URL="$arg"
+            ;;
+        --logs_enabled)
+            LOGS_ENABLED="$arg"
+            ;;
+        --metrics_enabled)
+            METRICS_ENABLED="$arg"
+            ;;
+        *)
+            echo "Unknown option: $opt"
+            exit 1
+            ;;
+    esac
+done
+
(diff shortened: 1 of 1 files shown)


scripts/install_mac.sh
```diff
@@ -40,17 +46,32 @@ sudo mkdir -p $observeagent_install_dir /usr/local/libexec /usr/local/bin /var/l
...
 sudo chown -R root:wheel $observeagent_install_dir
 
 # Initialize the agent config file if it doesn't exist
 if [ -f "$observeagent_install_dir/observe-agent.yaml" ]; then
     echo "Leaving existing observe-agent.yaml in place."
 else
     echo "Initializing observe-agent.yaml"
-    sudo $observeagent_install_dir/observe-agent init-config --token $TOKEN --observe_url $OBSERVE_URL --config_path $observeagent_install_dir/observe-agent.yaml
+    INIT_FLAGS="--config_path $observeagent_install_dir/observe-agent.yaml --token $TOKEN --observe_url $OBSERVE_URL --host_monitoring::enabled=true"
+    if [ -n "$LOGS_ENABLED" ]; then
```
obs-gh-alexlew, line 60: where are the keys for these args defined? is this just an env var? are we adding another layer/set of config key names?

obs-gh-mattcotter: Yes, these are just env vars. I needed some way of getting the args from the connection into the config. Now that you've called attention to this though, I wonder if it's better to switch this to `getopts` to be more in line with the windows script. What do you think about that?

obs-gh-mattcotter: Updated so it now works like:
`bash scripts/install_mac.sh --token=hi --observe_url example.com --logs_enabled=true`

scripts/install_mac.sh
```diff
@@ -4,9 +4,39 @@ set -e
 
 service_name="com.observeinc.agent"
 observeagent_install_dir="/usr/local/observe-agent"
-# TODO download the file instead of using the local built zip
-script_dir=$(dirname -- "$(readlink -f -- "$0")")
-zip_dir="$(dirname $script_dir)/dist/darwin_arm64_v8.0/observe-agent_Darwin_arm64.zip"
+tmp_dir="/tmp/observe-agent"
+
+# Parse args
+while [ $# -gt 0 ]; do
```
obs-gh-mattcotter, line 10: `getopts` doesn't work on mac for long options unfortunately, so I had to write out this approximation. It doesn't have all the bells and whistles, but is robust enough for our use case of having users copy/paste commands from the connection portal.

obs-gh-alexlew (approved):

PR #142, opened 2024-12-10
fix: clarify and expand the default env vars the agent provides to the configs

obs-gh-mattcotter:
### Description

Clarify and expand the default env vars the agent provides to the configs. This makes the vanilla collector endpoint available as well as exposing the prometheus endpoint. It also renames OBSERVE_TOKEN to OBSERVE_AUTHORIZATION_HEADER for clarity since the value is prefixed by `"Bearer "`, and renames OBSERVE_ENDPOINT to OBSERVE_OTEL_ENDPOINT since the value is suffixed with `"/v2/otel"` 

### Checklist

### DIFF
--- a/internal/connections/confighandler.go
+++ b/internal/connections/confighandler.go
@@ -78,13 +78,19 @@ func GetAllOtelConfigFilePaths(ctx context.Context, tmpDir string) ([]string, er
 
 func SetEnvVars() error {
 	collector_url, token, debug := viper.GetString("observe_url"), viper.GetString("token"), viper.GetBool("debug")
-	endpoint, err := url.JoinPath(collector_url, "/v2/otel")
+	otelEndpoint, err := url.JoinPath(collector_url, "/v2/otel")
+	if err != nil {
+		return err
+	}
+	promEndpoint, err := url.JoinPath(collector_url, "/v1/prometheus")
 	if err != nil {
 		return err
 	}
 	// Setting values from the Observe agent config as env vars to fill in the OTEL collector config
-	os.Setenv("OBSERVE_ENDPOINT", endpoint)
-	os.Setenv("OBSERVE_TOKEN", "Bearer "+token)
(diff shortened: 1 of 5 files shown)


internal/connections/confighandler.go
```diff
@@ -78,13 +78,19 @@ func GetAllOtelConfigFilePaths(ctx context.Context, tmpDir string) ([]string, er
...
+	promEndpoint, err := url.JoinPath(collector_url, "/v1/prometheus")
 	if err != nil {
 		return err
 	}
 	// Setting values from the Observe agent config as env vars to fill in the OTEL collector config
-	os.Setenv("OBSERVE_ENDPOINT", endpoint)
-	os.Setenv("OBSERVE_TOKEN", "Bearer "+token)
+	os.Setenv("OBSERVE_COLLECTOR_URL", collector_url)
+	os.Setenv("OBSERVE_OTEL_ENDPOINT", otelEndpoint)
+	os.Setenv("OBSERVE_PROMETHEUS_ENDPOINT", promEndpoint)
```
obs-gh-alexlew, line 92: where does this get used?

packaging/macos/config/otel-collector.yaml
```diff
@@ -83,9 +83,9 @@ processors:
 
 exporters:
   otlphttp/observe:
-    endpoint: ${env:OBSERVE_ENDPOINT}
+    endpoint: ${env:OBSERVE_OTEL_ENDPOINT}
```
obs-gh-alexlew, line 86: this works today and is just a cleanup?

obs-gh-mattcotter: Yes! This works, but it couldn't be used for prometheus or other collection as it had the `/v2/otel` baked into it. I think in an ideal world, users would specify the token and collection endpoint as top level env vars or config, and never reference them explicitly anywhere else.

PR #144, opened 2024-12-16
fix: update linux install to match mac install args

obs-gh-mattcotter:
### Description

Update linux install to match mac install args. Also only run the `systemd` setup when `systemd` is available.

### Checklist


obs-gh-alexlew:
have you tested this on a sandbox instance?

obs-gh-alexlew (approved):

obs-gh-mattcotter:
> have you tested this on a sandbox instance?

Yes! I didn't test on a box without systemd, but there seemed to be consensus online about the check that I copied to skip all of the systemd logic when it's not present

PR #146, opened 2024-12-18
feat: rename config flag to observe-config, use default otel command for our start command

obs-gh-mattcotter:
…for our start command

### Description

OB-38651 Rename `--config` flag to `--observe-config` and use the default OTel collector command for our start command. This adds all `otelcol` flags to our `start` command and gives us full functionality for the new `--config` flag (which specifies OTel configs).

### Checklist


main_windows.go
```diff
@@ -10,41 +10,44 @@ import (
...
+
+	// Get the collector settings along with our bundled config files.
+	configFilePaths, cleanup, err := start.SetupAndGetConfigFiles(start.DefaultLoggerCtx())
+	if cleanup != nil {
+		defer cleanup()
+	}
+	if err != nil {
+		return err
+	}
+	colSettings := observecol.GenerateCollectorSettingsWithConfigFiles(configFilePaths)
```
obs-gh-mattcotter, line 41: NB: Generating the settings this way instead of using the otel command means it won't work with `--config`. That's fine though, since we have a hard assert that the windows service is only ever run with two args and we have never supported `--config` this way.

internal/commands/start/start.go
```diff
@@ -42,46 +40,38 @@ func DefaultLoggerCtx() context.Context {
...
+		configFilePaths, cleanup, err := SetupAndGetConfigFiles(DefaultLoggerCtx())
+		if cleanup != nil {
+			defer cleanup()
+		}
 		if err != nil {
 			return err
 		}
-		if cleanup != nil {
-			defer cleanup()
+		configFlag := otleCmd.Flags().Lookup("config")
```
obs-gh-mattcotter, line 65: The flags for the otel collector are hidden behind private implementations of interfaces, args passed by value, and function scope locals. It's really not possible to access and set them directly if you aren't passing args to the cmd flags. Because of this, I think using the otelcol command itself for our `start` subcommand is the best way to go.

observecol/otelcollector.go
```diff
@@ -14,15 +14,7 @@ import (
 	"go.opentelemetry.io/collector/otelcol"
 )
 
-func makeMapProvidersMap(providers ...confmap.Provider) map[string]confmap.Provider {
```
obs-gh-mattcotter, line 17: This was unused.

internal/commands/diagnose/otelconfigcheck.go
```diff
@@ -16,13 +17,14 @@ type OtelConfigTestResult struct {
...
-		return false, nil, err
-	}
+	configFilePaths, cleanup, err := start.SetupAndGetConfigFiles(logger.WithCtx(context.Background(), logger.GetNop()))
 	if cleanup != nil {
 		defer cleanup()
 	}
+	if err != nil {
+		return false, nil, err
+	}
+	colSettings := observecol.GenerateCollectorSettingsWithConfigFiles(configFilePaths)
```
obs-gh-mattcotter, line 27: Diagnose won't run against otel config overrides, but we never supported this:
```
$ observe-agent diagnose --otel-config ./tmp.config
Error: unknown flag: --otel-config
Usage:
  observe-agent diagnose [flags]

Flags:
  -h, --help   help for diagnose

Global Flags:
      --config string   config file path
```

obs-gh-alexlew: can you make a ticket to support the new `--observe-config` flag in `diagnose`?

obs-gh-alexlew (approved):

internal/commands/diagnose/otelconfigcheck.go
obs-gh-mattcotter: Created https://observe.atlassian.net/browse/OB-40550

PR #161, opened 2025-02-06
feat: tune memory limiter settings

obs-gh-alexlew:
### Description

### Checklist

### DIFF
--- a/packaging/docker/observe-agent/otel-collector.yaml
+++ b/packaging/docker/observe-agent/otel-collector.yaml
@@ -33,8 +33,8 @@ processors:
 
   memory_limiter:
     check_interval: 1s
-    limit_percentage: 20
-    spike_limit_percentage: 5
+    limit_percentage: 80
+    spike_limit_percentage: 20
 
   batch:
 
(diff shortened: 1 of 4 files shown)


packaging/docker/observe-agent/otel-collector.yaml
```diff
@@ -33,8 +33,8 @@ processors:
 
   memory_limiter:
     check_interval: 1s
-    limit_percentage: 20
```
obs-gh-alexlew, line 36: I think we had misinterpreted the meaning of these fields. `limit_percentage` is the total possible usage allowed to this pipeline and after that it starts dropping data. I think maybe we thought that it wasn't a shared thing but actually each instance of this component has a shared size limit so if we only have one instance of this component then its safe to use up to 80% of memory total. The spike percentage is basically reserved for spikes so generally the process is actually limited to 60% of memory and is allowed to spike up to 80.

obs-gh-mattcotter: Nice, makes sense! So do you think we should omit this when running on bare metal? Maybe we could put a comment about that in our docs if so.

obs-gh-mattcotter (approved):

packaging/docker/observe-agent/otel-collector.yaml
obs-gh-alexlew: im not sure how the max gets calculated in a bare-metal situation since it doesn't technically exist. maybe there's a gomemlimit default? in which case i dont think its unreasonable to have this limiter try to respect that

PR #162, opened 2025-02-10
fix: send logs/agent-config over otlp as intended

obs-gh-mattcotter:
### Description

Send logs/agent-config over otlp as intended. This pipeline was mistakenly switched to prometheus.

### Checklist


packaging/macos/connections/self_monitoring/logs_and_metrics.yaml
```diff
@@ -29,4 +29,4 @@ service:
     logs/agent-config:
        receivers: [filelog/agent-config]
        processors: [memory_limiter, transform/truncate, resourcedetection, resourcedetection/cloud, batch]
-       exporters: [prometheusremotewrite/observe]
+       exporters: [otlphttp/observe]
```
obs-gh-mattcotter, line 32: Not sure how this happened. I verified that this pipeline in every other package folder is correctly OTLP.

PR #163, opened 2025-02-10
fix: ensure provided env var OBSERVE_COLLECTOR_URL never ends in a slash

obs-gh-mattcotter:
### Description

Ensure provided env var OBSERVE_COLLECTOR_URL never ends in a slash. Providing a consistent format regardless of how the agent is configured will make using this downstream easier and more reliable.

### Checklist

### DIFF
--- a/internal/connections/confighandler.go
+++ b/internal/connections/confighandler.go
@@ -74,6 +74,12 @@ func GetAllOtelConfigFilePaths(ctx context.Context, tmpDir string) ([]string, er
 
 func SetEnvVars() error {
 	collector_url, token, debug := viper.GetString("observe_url"), viper.GetString("token"), viper.GetBool("debug")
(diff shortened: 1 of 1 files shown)


internal/connections/confighandler.go
```diff
@@ -74,6 +74,10 @@ func GetAllOtelConfigFilePaths(ctx context.Context, tmpDir string) ([]string, er
 
 func SetEnvVars() error {
 	collector_url, token, debug := viper.GetString("observe_url"), viper.GetString("token"), viper.GetBool("debug")
+	// Ensure the collector url ends with a slash for consistency.
+	if collector_url[len(collector_url)-1] != '/' {
```
obs-gh-alexlew, line 78: it seems like its actually better to end without a slash generally and then use `url.joinPath` whenever we try to concat paths for the most consistent handling

obs-gh-alexlew (approved):

PR #131, opened 2024-11-13
ci: updates for windows 2019/2022 ami

obs-gh-nikhildua:
### Description

Updates AMIs for Windows 2019 / 2022 to new supported IDs

PR #164, opened 2025-02-10
fix: always re-add config values provided with --config after internal config values

obs-gh-mattcotter:
### Description

After switching to the otelcol's internal `--config` flag, the bundled config files provided by the `observe-agent` now are processed _after_ any configuration provided with `--config`. This makes it so the bundled config values cannot be overwritten via the `--config` flag (only via the `otel_config_overrides` option). This fix makes the flag take precedence again.

Ex:
```
$ cat ./hc_test.yaml 
extensions:
  health_check:
    endpoint: 0.0.0.0:13137
```

Current 2.0.0:
```
$ observe-agent start --config ./hc_test.yaml 
# ...
2025-02-10T11:10:54.862-0600	info	healthcheckextension/healthcheckextension.go:32	Starting health_check extension	{"kind": "extension", "name": "health_check", "config": {"Endpoint":"localhost:13133","TLSSetting":null,"CORS":null,"Auth":null,"MaxRequestBodySize":0,"IncludeMetadata":false,"ResponseHeaders":null,"CompressionAlgorithms":null,"ReadTimeout":0,"ReadHeaderTimeout":0,"WriteTimeout":0,"IdleTimeout":0,"Path":"/status","ResponseBody":null,"CheckCollectorPipeline":{"Enabled":false,"Interval":"5m","ExporterFailureThreshold":5}}}
```
^ note the `"Endpoint":"localhost:13133"`

After this fix:
```
$ ./observe-agent start --config ./hc_test.yaml 
# ...
2025-02-10T11:11:41.227-0600	info	healthcheckextension/healthcheckextension.go:32	Starting health_check extension	{"kind": "extension", "name": "health_check", "config": {"Endpoint":"0.0.0.0:13137","TLSSetting":null,"CORS":null,"Auth":null,"MaxRequestBodySize":0,"IncludeMetadata":false,"ResponseHeaders":null,"CompressionAlgorithms":null,"ReadTimeout":0,"ReadHeaderTimeout":0,"WriteTimeout":0,"IdleTimeout":0,"Path":"/status","ResponseBody":null,"CheckCollectorPipeline":{"Enabled":false,"Interval":"5m","ExporterFailureThreshold":5}}}
```
^ note the `"Endpoint":"0.0.0.0:13137"`

### Checklist

### DIFF
--- a/internal/commands/start/start.go
+++ b/internal/commands/start/start.go
@@ -6,6 +6,7 @@ package start
 import (
 	"context"
 	"os"
+	"strings"
 
 	logger "github.com/observeinc/observe-agent/internal/commands/util"
 	"github.com/observeinc/observe-agent/internal/connections"
@@ -55,7 +56,7 @@ collector on the current host.`
 	// Modify the run function so we can pass in our packaged config files.
 	originalRunE := otleCmd.RunE
 	otleCmd.RunE = func(cmd *cobra.Command, args []string) error {
-		configFilePaths, cleanup, err := SetupAndGetConfigFiles(DefaultLoggerCtx())
+		observeAgentConfigs, cleanup, err := SetupAndGetConfigFiles(DefaultLoggerCtx())
 		if cleanup != nil {
 			defer cleanup()
 		}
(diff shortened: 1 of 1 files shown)


integration/modules/create_ec2/locals.tf
```diff
@@ -115,7 +115,7 @@ locals {
   
      WINDOWS_SERVER_2022_BASE = {
       ami_instance_type = "t3.small"
-      ami_id            = "ami-00bedb8509ebcc120"
```
obs-gh-alexlew, line 118: why are these guys changing?

obs-gh-mattcotter: Terraform couldn't find the old AMIs. This happens from [time](https://github.com/observeinc/observe-agent/pull/125) to [time](https://github.com/observeinc/observe-agent/pull/131), but I don't know why.

internal/commands/start/start.go
```diff
@@ -55,15 +56,34 @@ collector on the current host.`
...
+		//
+		// However, we want any overrides that users provide via `--config` to take precedence, so we need to add our
+		// config values _before_ any current values. Since we can only append to the end of the flag, we add our
+		// bundled configs, then *re-add* the original config values.
+		originalConfigStr := configFlag.Value.String()
+		// This string is formatted as "[path1, path2, ...]"
+		// see: https://github.com/open-telemetry/opentelemetry-collector/blob/v0.118.0/otelcol/flags.go#L28-L30
+		// Trim the surrounding brackets, then split on ", "
+		var originalConfigs []string
+		if len(originalConfigStr) > 2 {
```
obs-gh-alexlew, line 79: this part is sort of confusing/seems unfortunate to me. are the first 2 arguments special in some way?

obs-gh-mattcotter: This is referring to the string; two characters only means `[]` which means no args. This skips the string split when we don't need to do it basically.

obs-gh-mattcotter: Oh and one more weird thing: calling Split on empty string does not return empty array but instead returns array of empty string.
https://go.dev/play/p/wZ3He29DP37

So this saves from filtering out empty string when looping over the configs below.

obs-gh-alexlew (approved):

PR #169, opened 2025-02-26
fix: make full use of nightly build functionality by publishing github releases and not tagging docker images as latest

obs-gh-mattcotter:
### Description

The objective of this change is to create github releases and publish docker images whenever we do a nightly build, but not tag either as `latest` (and instead use tag `dev-nightly`). Before this change: no github releases were published, and docker builds were published and tagged `latest`.

### Checklist

### DIFF
--- a/.github/workflows/release-nightly.yaml
+++ b/.github/workflows/release-nightly.yaml
@@ -23,6 +23,18 @@ jobs:
           fetch-depth: 0
           ref: ${{ github.event.inputs.branch }}
 
+      - name: Update the dev-nightly tag
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.git.updateRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'tags/dev-nightly',
+              sha: context.sha,
+              force: true
+            })
+
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
--- a/.github/workflows/release-nightly.yaml
+++ b/.github/workflows/release-nightly.yaml
@@ -40,9 +40,6 @@ jobs:
         with:
           go-version: 1.22.12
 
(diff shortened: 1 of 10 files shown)


obs-gh-mattcotter:
I believe that I can run this workflow before merging, but I will wait until the changes get approved to do so.

obs-gh-alexlew (approved):

obs-gh-mattcotter:
@obs-gh-alexlew  this seems to be working: https://github.com/observeinc/observe-agent/releases/tag/dev-nightly

obs-gh-mattcotter:
I verified also that docker tags look good:
https://hub.docker.com/r/observeinc/observe-agent/tags

.github/workflows/release-nightly.yaml
```diff
@@ -23,14 +23,23 @@ jobs:
           fetch-depth: 0
           ref: ${{ github.event.inputs.branch }}
 
+      - name: Update the dev-nightly tag
```
obs-gh-alexlew, line 26: why is this part necessary?

obs-gh-alexlew (approved):

.github/workflows/release-nightly.yaml
obs-gh-mattcotter: Github releases are [based on git tags](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#about-releases). I think we always need a git tag in order to create a GH release. This step updates the `dev-nightly` tag to the commit from which we are building the release. To keep our releases clean, I figured it would be nice to just have a single nightly tag (I believe this is how the [goreleaser repo handles nightly release](https://github.com/goreleaser/goreleaser-pro/releases/tag/nightly)).

obs-gh-alexlew (approved):

PR #171, opened 2025-03-05
internal: add CVE check to release job as blocking

obs-gh-alexlew:
### Description

internal: add CVE check to release job as blocking

### Checklist

### DIFF
--- a/.github/workflows/vuln-check.yaml
+++ b/.github/workflows/vuln-check.yaml
@@ -47,9 +47,6 @@ jobs:
         run: |
           cp dist/default_linux_386_sse2/observe-agent .
       
-      - name: Print files
-        run: ls
-
       - name: Build an image from Dockerfile
         run: docker build -f packaging/docker/Dockerfile -t docker.io/${{ env.TEST_TAG}} .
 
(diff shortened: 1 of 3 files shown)


.github/workflows/vuln-check.yaml
```diff
@@ -10,33 +11,45 @@ env:
...
-        uses: actions/upload-artifact@v4
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v6
         with:
-          name: myimage
-          path: ${{ runner.temp }}/myimage.tar
-  
+          distribution: goreleaser-pro
+          version: 2.4.4
+          args: build --snapshot --id=default --skip=validate
```
obs-gh-mattcotter, line 42: Does the binary we need get built with `--single-target`? If so, we can skip other builds to save some cycles.

obs-gh-mattcotter (approved):
LGTM!
This diff here fixes the go releaser failure fwiw:
https://github.com/observeinc/observe-agent/pull/172/files#diff-7326b55c062b0f46fe9e39aace0a25f4515cf206040fb91a6fd2cae839f5e826

.github/workflows/vuln-check.yaml
obs-gh-alexlew: ill test that out, should be only a single target required

PR #173, opened 2025-03-07
Add default package routing headers

obs-gh-jabernet:
In order for the out of box routing to work, send all traces to shared "Tracing" package and all other data to host connection specific "Host Explorer".

In the future those will be configurable.

### Description
https://observe.atlassian.net/browse/OB-41698


obs-gh-mattcotter (approved):
lgtm! I can't tell from the logs why the test was cancelled, might be worth rerunning.

obs-gh-mattcotter:
I rebased and tested on Mac and it looks good!

PR #175, opened 2025-03-13
fix: add explicit timeout to observe-agent integration tests

obs-gh-mattcotter:
### Description

Add explicit timeout to observe-agent integration tests

### DIFF
--- a/.github/workflows/tests-integration.yaml
+++ b/.github/workflows/tests-integration.yaml
@@ -8,20 +8,19 @@ on:
       - main
   schedule:
     - cron: "0 7 * * *" #Runs at 07:00 UTC Daily
-  
 
 #Special permissions required for OIDC authentication
 permissions:
-    id-token: write
-    contents: write
-    pull-requests: write
+  id-token: write
+  contents: write
+  pull-requests: write
 
 env:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests-integration.yaml
```diff
@@ -38,12 +37,13 @@ jobs:
       - name: Install qemu
         uses: docker/setup-qemu-action@v3
       - name: Run GoReleaser
+        timeout-minutes: 15
```
obs-gh-mattcotter, line 40: This is the only change in this file. There was a lot of trailing whitespace so I ran a yaml formatter and that accounts for all other diffs.

.goreleaser.yaml
```diff
@@ -40,12 +40,12 @@ builds:
...
     goarch:
       - amd64
-      - '386'
+      - "386"
     ldflags: -s -w -X github.com/observeinc/observe-agent/build.Version={{.Version}}
 
 archives:
   - id: default
-    format: tar.gz
+    formats: ["tar.gz"]
```
obs-gh-mattcotter, line 48: The old option was deprecated, this is the recommended new usage.
https://goreleaser.com/deprecations/#archivesformat

obs-gh-alexlew (approved):

PR #176, opened 2025-03-14
feat: add default batch timeout of 5s

obs-gh-mattcotter:
### Description

Add default batch timeout of 5s to reduce number of requests.

### Checklist

### DIFF
--- a/packaging/docker/observe-agent/otel-collector.yaml
+++ b/packaging/docker/observe-agent/otel-collector.yaml
@@ -37,6 +37,7 @@ processors:
     spike_limit_percentage: 20
 
   batch:
+    timeout: 5s
 
   deltatocumulative:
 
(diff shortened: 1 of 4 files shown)


obs-gh-jabernet:
https://docs.observeinc.com/en/latest/content/data-ingestion/endpoints/otel.html ; Our per request limit is 50mb uncompressed. So we could have significantly higher thresholds assuming 1kb log lines. Maybe we should just set both values to the default `send_batch_size` of 8192 for now and we can tune it in another PR based on data from O2?

obs-gh-mattcotter:
> https://docs.observeinc.com/en/latest/content/data-ingestion/endpoints/otel.html ; Our per request limit is 50mb uncompressed. So we could have significantly higher thresholds assuming 1kb log lines. Maybe we should just set both values to the default `send_batch_size` of 8192 for now and we can tune it in another PR based on data from O2?

I have updated this PR to only set the timeout and leave the sizing values as default. Thanks!

obs-gh-alexlew (approved):

PR #179, opened 2025-03-20
fix: only block release on fixed issues with higher severity than low

obs-gh-mattcotter:
### Description

Only block release on fixed issues with higher severity than low. This will ensure we can still release in the 180 day window we have to fix low severity issues, or when we know about an issue that has no fix.

### DIFF
--- a/.github/workflows/vuln-check-full.yaml
+++ b/.github/workflows/vuln-check-full.yaml
@@ -7,11 +7,6 @@ on:
         description: "The branch on which to run the vulnerability check"
         required: true
         default: "main"
-  workflow_call:
-    inputs:
-      branch:
-        required: true
-        type: string
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


obs-gh-alexlew (approved):

.github/workflows/vuln-check-full.yaml
```diff
@@ -7,11 +7,6 @@ on:
         description: "The branch on which to run the vulnerability check"
         required: true
         default: "main"
-  workflow_call:
```
obs-gh-alexlew, line 10: why are we removing this from the full check? I'd still want to run this against an arbitrary branch

obs-gh-mattcotter: Manual runs go through the `workflow_dispatch` (https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch), which still has the branch input above.

If we think we would call this full run workflow as a step in a different workflow, then I can leave in the `workflow_call`. I didn't foresee that happening though

PR #182, opened 2025-03-27
feat: add agent config options for internal telemetry and health check, enable use of status cmd in kubernetes

obs-gh-mattcotter:
### Description

Add agent config options for internal telemetry and health check. Enable the use of the agent `status` command in kubernetes. Also upgrade the internal telemetry config as our config for metrics is no longer supported: https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.123.0

Ex:
```
$ kubectl -n observe exec --stdin --tty observe-agent-node-logs-metrics-agent-brjwr -- /observe-agent --observe-config=/observe-agent-conf/observe-agent.yaml status
Defaulted container "node-logs-metrics" out of: node-logs-metrics, kube-cluster-info (init)
================
Agent
================

  Host Info
  ================
  HostID: 45220b39-2c66-47bf-984d-9d0bbae88222
  Hostname: observe-agent-node-logs-metrics-agent-brjwr
  BootTime: 2025-04-10T20:05:46Z
  Uptime: 20h32m12s
  OS: linux
  Platform: alpine
  PlatformFamily: alpine
  PlatformVersion: 3.21.3
  KernelArch: aarch64
  KernelVersion: 6.10.14-linuxkit

  Agent Metrics
  ================
  ExporterQueueSize: 0
  CPUSeconds: 1.19s
  MemoryUsed: 132.86328MB
  TotalSysMemory: 60.27076MB
  Uptime: 17.37295s
  AvgServerResponseTime: 0ms
  AvgClientResponseTime: 0ms

    Logs Stats
    ================
    ReceiverAcceptedCount: 18
    ReceiverRefusedCount: 0
    ExporterSentCount: 10
    ExporterSendFailedCount: 0

    Metrics Stats
    ================
    ReceiverAcceptedCount: 601
    ReceiverRefusedCount: 0
    ExporterSentCount: 601
    ExporterSendFailedCount: 0

    Traces Stats
    ================
    ReceiverAcceptedCount: 0
    ReceiverRefusedCount: 0
    ExporterSentCount: 0
    ExporterSendFailedCount: 0

  Agent Health
  ================
  Status: Running
  TotalRefusedCount: 0
  TotalSendFailedCount: 0
```


internal/commands/status/flags.go
```diff
@@ -0,0 +1,30 @@
+package status
+
+import (
+	"github.com/spf13/viper"
+
+	"github.com/spf13/cobra"
+)
+
+const (
+	TelemetryEndpointFlag   = "telemetry-endpoint"
```
obs-gh-alexlew, line 10: I think actually the more useful use case here is that we could then provide overrides in the `config.yaml` we generate in the helm chart. so we could add a `telemetry_endpoint: "{{ template "config.local_host"}}:8888"` and that would work correctly. Since that's the case, I'd prefer if we stick to the snake_case style of naming instead of kebabcase and maybe we can nest these under something like `endpoints::telemetry_endpoint`

obs-gh-mattcotter: Good point, I agree! I will update this PR after my config refactor to simplify the rebasing.

internal/commands/config/config.go
```diff
@@ -49,96 +41,14 @@ bundled OTel configuration.`,
...
 		} else if detailedOtel {
-			return printFullOtelConfig(configFilePaths)
+			return util.PrintFullOtelConfig(configFilePaths)
 		}
-		return printAllConfigsIndividually(configFilePaths)
+		return util.PrintAllConfigsIndividually(configFilePaths)
 	},
 }
 
-func printAllConfigsIndividually(configFilePaths []string) error {
```
obs-gh-mattcotter, line 60: This is all code motion; I put these methods in `util` since the config package depends on the start package, and having these methods available in `start` for debugging is very useful.

internal/commands/status/statusretriever.go
```diff
@@ -68,16 +72,26 @@ func bToMb(b float32) float32 {
 	return b / 1024 / 1024
 }
 
-func GetAgentStatusFromHealthcheck(baseURL string) (AgentStatus, error) {
-	URL := fmt.Sprintf("%s/status", baseURL)
+func GetAgentStatusFromHealthcheck(baseURL string, path string) (AgentStatus, error) {
+	baseURL = util.ReplaceEnvString(baseURL)
```
obs-gh-mattcotter, line 76: This is to handle our default k8s use of `${env:MY_POD_IP}`

internal/commands/status/statusretriever.go
```diff
@@ -119,53 +147,53 @@ func GetAgentMetricsFromEndpoint(baseURL string) (*AgentMetrics, error) {
 		if v.Type.String() == io_prometheus_client.MetricType_HISTOGRAM.String() {
 			met := v.Metric[0]
 			switch name := *v.Name; name {
-			case "otelcol_http_client_duration":
+			case "http_client_duration_milliseconds":
```
obs-gh-mattcotter, line 150: These metric names changed; possibly to be inline with prometheus conventions after the config upgrade. I checked the output from the prometheus endpoint and verified the metric names as well as watching the numbers update when calling the status command.

internal/commands/util/config_printers.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/go-viper/mapstructure/v2"
+	"github.com/observeinc/observe-agent/internal/config"
+	"github.com/observeinc/observe-agent/observecol"
+	"github.com/spf13/viper"
+	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/otelcol"
+	"gopkg.in/yaml.v3"
+)
+
+func PrintAllConfigsIndividually(configFilePaths []string) error {
```
obs-gh-mattcotter, line 18: Here's where the config methods moved to. Again, no changes just code motion.

internal/commands/status/statusretriever.go
obs-gh-mattcotter: I verified that the names after this update match what's collected now from our prometheus exporter

obs-gh-alexlew (approved):

PR #184, opened 2025-04-02
feat: add config command flag --render-otel to print full otel config

obs-gh-mattcotter:
### Description

Add the flags `--render-otel` and `--render-otel-details` to the `config` command to print full otel config. This short version is modeled after the `otelcol print` command, the long version was modeled after the startup command.

Ex:
```
$ ./observe-agent config --render-otel | tail -n 40
                - debug
            processors:
                - filter/count
            receivers:
                - count
        metrics/forward:
            exporters:
                - prometheusremotewrite/observe
            processors:
                - resourcedetection
                - resourcedetection/cloud
                - deltatocumulative
                - batch
            receivers:
                - otlp
        metrics/host_monitoring_host:
            exporters:
                - prometheusremotewrite/observe
            processors:
                - memory_limiter
                - resourcedetection
                - resourcedetection/cloud
                - batch
            receivers:
                - hostmetrics/host-monitoring-host
        traces/forward:
            exporters:
                - otlphttp/observetracing
            processors:
                - resourcedetection
                - resourcedetection/cloud
            receivers:
                - otlp
    telemetry:
        logs:
            level: INFO
        metrics:
            address: :8888
            level: detailed
```
vs
```
$ ./observe-agent config --render-otel-details | tail -n 40
            receivers:
                - otlp
            processors:
                - resourcedetection
                - resourcedetection/cloud
            exporters:
                - otlphttp/observe
                - count
    telemetry:
        logs:
            encoding: console
            error_output_paths:
                - stderr
            level: info
            output_paths:
                - stderr
            sampling:
                enabled: true
                initial: 10
                thereafter: 100
                tick: 10s
        metrics:
            address: :8888
            level: Detailed
            readers:
                - pull:
                    exporter:
                        prometheus:
                            host: ""
                            port: 8888
                            with_resource_constant_labels: {}
                            without_scope_info: true
                            without_type_suffix: true
                            without_units: true
                        additionalproperties: null
        traces:
            level: Basic
            processors: []
            propagators: []
```

### DIFF
--- a/internal/commands/config/config.go
+++ b/internal/commands/config/config.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/go-viper/mapstructure/v2"
 	"github.com/observeinc/observe-agent/internal/commands/start"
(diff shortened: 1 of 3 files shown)


obs-gh-alexlew:
how is this different from the `config` command?

obs-gh-mattcotter:
> how is this different from the `config` command? or you're looking to deprecate that one

Good question! The `config` command lists out all fragments individually and also dumps the observe-agent config. Really the original mode plus the two others here all serve different purposes. I don't know if we actually want to support three ways of spitting out configs, but the two here are definitely useful and the existing one has been useful in the past. Since it's not much code I'm fine with having all three for now, but let me know if you feel differently.

internal/commands/config/config.go
```diff
@@ -7,57 +7,134 @@ import (
...
+	}
+	fmt.Printf("%s\n", b)
+	return nil
+}
+
+func printFullOtelConfig(configFilePaths []string) error {
+	colSettings := observecol.GenerateCollectorSettingsWithConfigFiles(configFilePaths)
+	factories, err := colSettings.Factories()
+	if err != nil {
+		return fmt.Errorf("failed to initialize factories: %w", err)
```
obs-gh-alexlew, line 113: Nit: are we initializing or just retrieving the existing factories?

obs-gh-alexlew (approved):

internal/commands/config/config.go
obs-gh-mattcotter: Good point, we're initializing the maps. I rephrased this to say: "failed to create component factory maps"

PR #186, opened 2025-04-07
refactor: split main otel config into multiple connections, add tests

obs-gh-mattcotter:
### Description

OB-36266: Split main otel config into multiple connections. Use the agent config object as the source of truth instead of viper. Add tests.

I also tested this with the new `observe-agent config --render-otel` command. I enabled all options in my agent config, and diffed the output between `main` and this branch and only these lines changed:
```diff
104c104
<             - /usr/local/observe-agent/config/otel-collector.yaml
---
>             - /usr/local/observe-agent/connections/common/base.yaml.tmpl
123c123
<         include: /usr/local/observe-agent/config/otel-collector.yaml
---
>         include: /usr/local/observe-agent/connections/common/base.yaml.tmpl
```
This is expected since the otel-collector.yaml file no longer exists.

### Checklist


internal/commands/initconfig/initconfig_test.go
```diff
@@ -77,17 +78,31 @@ func Test_InitConfigCommand(t *testing.T) {
...
-		}
-		err = yaml.Unmarshal(configFile, &config)
-		if err != nil {
-			t.Errorf("Expected no error, got %v", err)
-		}
-		assert.Equal(t, tc.expectedConfig, config)
+
+		var configYaml, configMapstructure config.AgentConfig
+
+		// Decode via mapstrucutre (which is how viper does it)
```
obs-gh-alexlew, line 88: nit typo mapstructure

internal/connections/allconnectiontypes.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+var CommonConnectionType = MakeConnectionType(
+	"common",
+	func(_ *config.AgentConfig) bool {
+		return true
+	},
+	[]CollectorConfigFragment{
+		{
+			enabledCheck: func(agentConfig *config.AgentConfig) bool {
```
obs-gh-alexlew, line 20: these are basically the shared components between other connection types right? so if we add more connection types we'd need to add them to this check here too? any reason to not just always include this?

internal/connections/allconnectiontypes.go
```diff
@@ -0,0 +1,89 @@
...
+				return (agentConfig.Forwarding.Enabled ||
+					agentConfig.HealthCheck.Enabled ||
+					agentConfig.SelfMonitoring.Enabled ||
+					agentConfig.HostMonitoring.Enabled)
+			},
+			colConfigFilePath: "base.yaml.tmpl",
+		},
+		{
+			enabledCheck: func(agentConfig *config.AgentConfig) bool {
+				return agentConfig.Forwarding.Enabled
```
obs-gh-alexlew, line 30: before there was no value here right? so if we change this to be default off, that's a breaking change I think. Same for the healthcheck below. do these default to on?

obs-gh-mattcotter: These do default to on, but it happens via defaulting the Enabled variable to true in the config definition. I added "default" struct tags to `config.AgentConfig` and the `go-defaults` library to set them. The new `TestAgentConfigFromViper` validates that this works as intended.

internal/connections/allconnectiontypes.go
obs-gh-mattcotter: That's right. I was thinking it would be nice to have a total noop mode (ie operate exactly like a vanilla `otelcol`), but it's probably not worth the extra complexity here. I changed this to always return true.

obs-gh-alexlew (approved):

PR #188, opened 2025-04-17
fix: add defaults for agent config to fully support nested env var overriders

obs-gh-mattcotter:
### Description

Add viper defaults for all agent config settings. Without this, nested env vars will be ignored unless the value is already set in the config. ex: `env HOST_MONITORING::ENABLED=true ./observe-agent start` will be ignored with an empty agent config.

### DIFF
--- a/internal/config/configschema.go
+++ b/internal/config/configschema.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/go-viper/mapstructure/v2"
 	"github.com/mcuadros/go-defaults"
 	"github.com/spf13/viper"
 )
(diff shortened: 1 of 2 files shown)


internal/config/configschema.go
```diff
@@ -68,6 +69,32 @@ type AgentConfig struct {
...
 
+func SetViperDefaults(v *viper.Viper) {
+	var config AgentConfig
+	defaults.SetDefaults(&config)
+	var confMap map[string]any
+	err := mapstructure.Decode(config, &confMap)
+	if err != nil {
+		panic(err)
+	}
+	var recursiveDfs func(prefix string, defaults map[string]any)
```
obs-gh-alexlew, line 80: why not just `recursiveDfs := func(`?

internal/root/root.go
```diff
@@ -53,6 +54,11 @@ func InitConfig() {
 	}
 
 	viper.AutomaticEnv() // read in environment variables that match
+	// TODO add this into our next major release to improve our env var naming.
+	// viper.SetEnvKeyReplacer(strings.NewReplacer(`::`, `_`))
```
obs-gh-alexlew, line 58: we use _ in the actual field names as well, would this work with that? that's why we switched to `::` to begin with

obs-gh-alexlew (approved):

internal/config/configschema.go
obs-gh-mattcotter: I need to declare it first since it's recursive, otherwise I can't reference it in the function body.

internal/root/root.go
obs-gh-mattcotter: Oh, if the `::` was to avoid ambiguity, then I will just delete this comment. I think that `_` is easier to use (ie `HOST_MONITORING_ENABLED=true observe-agent start` vs `env HOST_MONITORING::ENABLED=true observe-agent start`), but if this was an intentional choice for clarity then let's keep it.

PR #198, opened 2025-04-23
feat: default the TRACE_TOKEN env var to be the agent token

obs-gh-mattcotter:
OB-42780: default the TRACE_TOKEN env var to be the agent token

I think this is the simplest solution and avoids a lot of k8s complexity.

### DIFF
--- a/internal/connections/confighandler.go
+++ b/internal/connections/confighandler.go
@@ -77,6 +77,13 @@ func SetEnvVars() error {
 	os.Setenv("OBSERVE_AUTHORIZATION_HEADER", "Bearer "+token)
 	os.Setenv("FILESTORAGE_PATH", GetDefaultFilestoragePath())
 
+	// Default TRACE_TOKEN to be the value of the configured token if it's not set. This allows for users to upgrade to
+	// direct write tracing with ingest tokens in kubernetes without breaking backwards compatibility in our helm chart.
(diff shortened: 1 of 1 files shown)


obs-gh-jabernet (approved):
LGTM, seems most pragmatic way to approach this for now. Maybe add a TODO to remove this once we remove the `TRACE_TOKEN` branching in helm chart.

internal/connections/confighandler.go
```diff
@@ -77,6 +77,12 @@ func SetEnvVars() error {
 	os.Setenv("OBSERVE_AUTHORIZATION_HEADER", "Bearer "+token)
 	os.Setenv("FILESTORAGE_PATH", GetDefaultFilestoragePath())
 
+	// Default TRACE_TOKEN to be the value of the configured token if it's not set. This allows for users to upgrade to
+	// direct write tracing with ingest tokens in kubernetes without breaking backwards compatibility in our helm chart.
+	if os.Getenv("TRACE_TOKEN") == "" {
```
obs-gh-alexlew, line 82: when would we want to remove this?

obs-gh-mattcotter: I think we will want this as long as our helm chart supports `TRACE_TOKEN`. After that it should be fine to remove this logic. I added that as a TODO as Janick suggested

PR #202, opened 2025-05-06
feat: add config option to set service explorer attributes

obs-gh-mattcotter:
### Description

OB-44816 Add config option to set generic attributes or resource attributes. This is needed to configure service explorer attributes when running on a host.

### Checklist

### DIFF
--- a/integration/scripts/test_configure.py
+++ b/integration/scripts/test_configure.py
@@ -2,44 +2,47 @@
 import utils as u
 
 
-@u.print_test_decorator
-def run_test_windows(remote_host: u.Host, env_vars: dict) -> None:
-    init_command = r'Set-Location "C:\Program Files\Observe\observe-agent"; ./observe-agent init-config --token {} --observe_url {} --cloud_resource_detectors ec2'.format(
+def init_config_command(env_vars: dict) -> str:
+    return "init-config --token {} --observe_url {} --cloud_resource_detectors ec2 --resource_attributes deployment.environment=test".format(
         env_vars["observe_token"], env_vars["observe_url"]
     )
+
+
+def run_init_config_common(remote_host: u.Host, init_command: str) -> None:
+    # Print the config to be used first
+    result = remote_host.run_command(init_command + " --print")
+    print("Setting agent config:\n{}\n".format("=" * 21))
+    u.print_remote_result(result)
+    if result.exited != 0 or result.stderr:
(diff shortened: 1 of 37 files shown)


integration/scripts/test_configure.py
```diff
@@ -2,12 +2,20 @@
 import utils as u
 
 
-@u.print_test_decorator
-def run_test_windows(remote_host: u.Host, env_vars: dict) -> None:
-    init_command = r'Set-Location "C:\Program Files\Observe\observe-agent"; ./observe-agent init-config --token {} --observe_url {} --cloud_resource_detectors ec2'.format(
+def init_config_command(env_vars: dict) -> str:
```
obs-gh-mattcotter, line 5: This change sets a resource attribute in the config for all tests. I had some yaml spacing issues in my diff originally, so I added extra logging; that is why the test files have other changes.

integration/scripts/test_start.py
```diff
@@ -65,7 +66,7 @@ def run_test_windows(remote_host: u.Host, env_vars: dict) -> None:
     # status
     start_command = r".\start_agent_windows.ps1"
     status_command = r'Get-Service ObserveAgent;Set-Location "${Env:Programfiles}\Observe\observe-agent"; ./observe-agent status'
-    start_timeout = 30  # how long to wait for observe-agent to start
+    start_timeout = 10  # how many times to check for the running agent
```
obs-gh-alexlew, line 69: is this how many times we check? is start_timeout not a correct variable name?

packaging/docker/observe-agent/connections/common/attributes.yaml.tmpl
```diff
@@ -0,0 +1,19 @@
+processors:
+  {{- if .HasAttributes }}
+  attributes/observe_global_attributes:
+    actions:
+    {{- range $key, $value := .Attributes }}
+      - key: {{ $key }}
+        value: {{ $value }}
+        action: insert
```
obs-gh-alexlew, line 8: i think maybe this might need to be overrideable as well, upsert is potentially a desired value

integration/scripts/test_start.py
obs-gh-mattcotter: Good point, I renamed this and added/used a default value at the call sites.

obs-gh-alexlew (approved):

PR #206, opened 2025-05-09
feat: add snapshot tests for bundled config

obs-gh-mattcotter:
### Description

OB-45143 Add snapshot tests for the full bundled config in our test environments. This will catch when the observe-agent's provided otel config changes.

### Checklist

### DIFF
--- a/integration/scripts/snapshots/docker.yaml
+++ b/integration/scripts/snapshots/docker.yaml
@@ -1,0 +1,366 @@
+connectors:
+    count: null
+exporters:
+    debug: null
+    nop: null
+    otlphttp/observe:
+        compression: zstd
+        endpoint: https://123456789.collect.observe-eng.com/v2/otel
+        headers:
+            authorization: Bearer abcdefghijklmnopqrst:OWt0SXV4YTlqYUhpSHZjSlhXUDVkRXpl
+            x-observe-target-package: Host Explorer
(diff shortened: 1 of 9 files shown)


integration/scripts/snapshots/docker.yaml
```diff
@@ -0,0 +1,366 @@
+connectors:
+    count: null
+exporters:
+    debug: null
+    nop: null
+    otlphttp/observe:
+        compression: zstd
+        endpoint: https://123456789.collect.observe-eng.com/v2/otel
+        headers:
+            authorization: Bearer abcdefghijklmnopqrst:OWt0SXV4YTlqYUhpSHZjSlhXUDVkRXpl
```
obs-gh-alexlew, line 10: is there some way to mark to orca that this isn't a real credential? i feel like it might complain although the tests seem to pass now?

obs-gh-alexlew:
how did the snapshots get generated currently? and when am I expected to update these?

integration/scripts/snapshots/docker.yaml
obs-gh-mattcotter: @obs-gh-nikhildua do you know if this is possible? Maybe we can figure that out if it ever does complain?

obs-gh-alexlew (approved):

PR #207, opened 2025-05-12
add feature flag and parallelism config to PRWE

obs-gh-alexlew:
### Description

### Checklist

### DIFF
--- a/internal/commands/start/start.go
+++ b/internal/commands/start/start.go
@@ -86,6 +86,9 @@ collector on the current host.`
 		for _, path := range originalConfigs {
 			configFlag.Value.Set(path)
 		}
+		// Set PRWE enable multiple workers feature gate.
+		EnableMultipleWritersFlag := otleCmd.Flags().Lookup("feature-gates")
+		EnableMultipleWritersFlag.Value.Set("+exporter.prometheusremotewritexporter.EnableMultipleWorkers")
 		return originalRunE(cmd, args)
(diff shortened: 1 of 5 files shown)


internal/commands/start/start.go
```diff
@@ -86,6 +86,9 @@ collector on the current host.`
 		for _, path := range originalConfigs {
 			configFlag.Value.Set(path)
 		}
+		// Set PRWE enable multiple workers feature gate.
+		EnableMultipleWritersFlag := otleCmd.Flags().Lookup("feature-gates")
```
obs-gh-mattcotter, line 90: Is there a reason to start the name with an upper case E? I think lower case would be more conventional.

packaging/docker/observe-agent/connections/common/base.yaml.tmpl
```diff
@@ -97,6 +97,9 @@ exporters:
     headers:
       authorization: ${env:OBSERVE_AUTHORIZATION_HEADER}
       x-observe-target-package: "Host Explorer"
+    remote_write_queue:
+      num_consumers: 10
+    max_batch_request_parallelism: 10
```
obs-gh-mattcotter, line 102: This is set to 10 but the helm chart uses 5 for both of these values. I think it's better to be consistent unless I'm missing something

obs-gh-mattcotter (approved):

PR #212, opened 2025-05-19
feat: add option to send forwarded metrics to Observe via OTLP

obs-gh-mattcotter:
### Description

Add option to send forwarded metrics to Observe via OTLP. Going forward, we want to be able to send all custom metrics to Observe in the format the user prefers (between OTLP or Prometheus)

### DIFF
--- a/internal/commands/initconfig/initconfig.go
+++ b/internal/commands/initconfig/initconfig.go
@@ -19,7 +19,7 @@ var (
 	observe_url                             string
 	cloud_resource_detectors                []string
 	resource_attributes                     map[string]string
-	forwarding_use_otlp                     bool
+	forwarding_metrics_format               string
 	self_monitoring_enabled                 bool
 	host_monitoring_enabled                 bool
 	host_monitoring_logs_enabled            bool
@@ -88,7 +88,7 @@ func RegisterConfigFlags(cmd *cobra.Command, v *viper.Viper) {
 	cmd.PersistentFlags().StringToStringVar(&resource_attributes, "resource_attributes", map[string]string{}, "The cloud environments from which to detect resources")
 	v.BindPFlag("resource_attributes", cmd.PersistentFlags().Lookup("resource_attributes"))
 
-	cmd.PersistentFlags().BoolVar(&self_monitoring_enabled, "forwarding::metrics::output_format", false, "Format for sending app metrics to Observe, valid options are 'prometheus' and 'otel'")
+	cmd.PersistentFlags().StringVar(&forwarding_metrics_format, "forwarding::metrics::output_format", "", "Format for sending app metrics to Observe, valid options are 'prometheus' and 'otel'")
 	v.BindPFlag("forwarding::metrics::output_format", cmd.PersistentFlags().Lookup("forwarding::metrics::output_format"))
 
 	cmd.PersistentFlags().BoolVar(&self_monitoring_enabled, "self_monitoring::enabled", true, "Enable self monitoring")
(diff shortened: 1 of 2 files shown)


packaging/docker/observe-agent/connections/common/forward.yaml.tmpl
```diff
@@ -19,11 +19,24 @@ exporters:
       enabled: true
     compression: zstd
 
+  otlphttp/observemetrics:
+    endpoint: ${env:OBSERVE_OTEL_ENDPOINT}
+    headers:
+      authorization: ${env:OBSERVE_AUTHORIZATION_HEADER}
+      x-observe-target-package: "Metrics"
+    sending_queue:
+      num_consumers: 4
```
obs-gh-alexlew, line 28: how'd you land on these values?

packaging/linux/config/observe-agent.yaml
```diff
@@ -7,6 +7,11 @@ observe_url: "<OBSERVE COLLECTION ENDPOINT>"
 # Debug mode - Sets agent log level to debug
 debug: false
 
+# Enable forwarding of local app metrics and traces
+forwarding:
+  enabled: true
+  otlp_metrics: true
```
obs-gh-alexlew, line 13: its a little unclear that these are referring to the output format; maybe we should nest this in another object something like

```
forwarding:
  enabled: true
  output:
    metrics_format: "prometheus" or "OTLP"
```

packaging/docker/observe-agent/connections/common/forward.yaml.tmpl
obs-gh-mattcotter: These are the same values used in all of our otlphttp exporters. It's lower than the default, but I don't know how we arrived at them in the first place. Do you think we should make them configurable in the agent config?

packaging/linux/config/observe-agent.yaml
obs-gh-mattcotter: I went with:
```yaml
forwarding:
  enabled: true
  metrics:
    output_format: "otlp"
```
to match your helm change.

packaging/docker/observe-agent/connections/common/forward.yaml.tmpl
obs-gh-alexlew: dont need to in this PR

packaging/linux/config/observe-agent.yaml
obs-gh-alexlew: should the format be `otel` or `otlp`? the latter is more specific but references the protocol vs. the actual schema/data format

internal/commands/diagnose/configcheck.go
```diff
@@ -18,31 +19,52 @@ type ConfigTestResult struct {
 }
 
 func checkConfig(v *viper.Viper) (bool, any, error) {
+	// Ensure there is an observe-agent config file.
 	configFile := v.ConfigFileUsed()
 	if configFile == "" {
 		return false, nil, fmt.Errorf("no config file defined")
 	}
+	if _, err := os.Stat(configFile); err != nil && errors.Is(err, os.ErrNotExist) {
```
obs-gh-alexlew, line 27: this seems like an unrelated change? should it be in the same PR?

obs-gh-mattcotter: The diagnose check wasn't previously applying our config defaulting logic. Since this change adds a config option with a default that's invalid when empty, diagnose needed to be updated. So it's a loosely related bugfix that needs to happen before this option is added. I can split it out if you prefer?

obs-gh-alexlew (approved):

PR #214, opened 2025-05-29
feat: update pod status logic

obs-gh-alexlew:
### Description

### Checklist

### DIFF
--- a/components/processors/observek8sattributesprocessor/podactions.go
+++ b/components/processors/observek8sattributesprocessor/podactions.go
@@ -76,11 +76,6 @@ func NewPodStatusAction() PodStatusAction {
 // Generates the Pod "status" facet.
 func (PodStatusAction) ComputeAttributes(pod v1.Pod) (attributes, error) {
 	// based on https://github.com/kubernetes/kubernetes/blob/cd3b5c57668a0a6e32057ef82dfab40e9b0bec5b/pkg/printers/internalversion/printers.go#L881
-	restarts := 0
-	restartableInitContainerRestarts := 0
-	totalContainers := len(pod.Spec.Containers)
-	readyContainers := 0
-
 	podPhase := pod.Status.Phase
 	reason := string(podPhase)
 	if pod.Status.Reason != "" {
@@ -97,23 +92,16 @@ func (PodStatusAction) ComputeAttributes(pod v1.Pod) (attributes, error) {
 	initContainers := make(map[string]*v1.Container)
 	for i := range pod.Spec.InitContainers {
 		initContainers[pod.Spec.InitContainers[i].Name] = &pod.Spec.InitContainers[i]
-		if isRestartableInitContainer(&pod.Spec.InitContainers[i]) {
-			totalContainers++
-		}
 	}
 
 	initializing := false
 	for i := range pod.Status.InitContainerStatuses {
 		container := pod.Status.InitContainerStatuses[i]
-		restarts += int(container.RestartCount)
 		switch {
 		case container.State.Terminated != nil && container.State.Terminated.ExitCode == 0:
 			continue
 		case isRestartableInitContainer(initContainers[container.Name]) &&
 			container.Started != nil && *container.Started:
-			if container.Ready {
-				readyContainers++
-			}
 			continue
 		case container.State.Terminated != nil:
 			// initialization is failed
@@ -138,11 +126,9 @@ func (PodStatusAction) ComputeAttributes(pod v1.Pod) (attributes, error) {
 	}
 
 	if !initializing || isPodInitializedConditionTrue(&pod.Status) {
-		restarts = restartableInitContainerRestarts
 		hasRunning := false
 		for i := len(pod.Status.ContainerStatuses) - 1; i >= 0; i-- {
 			container := pod.Status.ContainerStatuses[i]
-			restarts += int(container.RestartCount)
 			if container.State.Waiting != nil && container.State.Waiting.Reason != "" {
 				reason = container.State.Waiting.Reason
 			} else if container.State.Terminated != nil && container.State.Terminated.Reason != "" {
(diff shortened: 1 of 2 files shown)


obs-gh-alexlew:
datadog code: https://github.com/DataDog/datadog-agent/blob/7d47d25899c025c8b3e5ba91e488fa0f97f7a275/pkg/collector/corechecks/cluster/orchestrator/transformers/k8s/pod.go#L221

obs-gh-alexlew:
tested in my sandbox instance and things look reasonable

components/processors/observek8sattributesprocessor/podactions.go
```diff
@@ -33,6 +35,38 @@ const (
...
+	return false
+}
+
+func isPodInitializedConditionTrue(status *v1.PodStatus) bool {
+	for _, condition := range status.Conditions {
+		if condition.Type != v1.PodInitialized {
+			continue
+		}
+
+		return condition.Status == v1.ConditionTrue
```
obs-gh-mattcotter, line 54: Maybe change this to 
```suggestion
		if condition.Type == v1.PodInitialized && condition.Status == v1.ConditionTrue {
			return true
		}
```
to match `hasPodReadyCondition`

components/processors/observek8sattributesprocessor/podactions.go
```diff
@@ -43,18 +77,63 @@ func NewPodStatusAction() PodStatusAction {
...
+		}
 		switch {
 		case container.State.Terminated != nil && container.State.Terminated.ExitCode == 0:
 			continue
+		case isRestartableInitContainer(initContainers[container.Name]) &&
+			container.Started != nil && *container.Started:
+			if container.Ready {
+				readyContainers++
+			}
+			continue
```
obs-gh-mattcotter, line 136: Should this be in the `switch` statement? It could skip the `default` case and miss setting the reason there.

components/processors/observek8sattributesprocessor/podactions.go
```diff
@@ -43,18 +77,63 @@ func NewPodStatusAction() PodStatusAction {
...
 func (PodStatusAction) ComputeAttributes(pod v1.Pod) (attributes, error) {
-	// based on https://github.com/kubernetes/kubernetes/blob/0d3b859af81e6a5f869a7766c8d45afd1c600b04/pkg/printers/internalversion/printers.go#L901
-	reason := string(pod.Status.Phase)
+	// based on https://github.com/kubernetes/kubernetes/blob/cd3b5c57668a0a6e32057ef82dfab40e9b0bec5b/pkg/printers/internalversion/printers.go#L881
+	restarts := 0
+	restartableInitContainerRestarts := 0
+	totalContainers := len(pod.Spec.Containers)
+	readyContainers := 0
+	lastRestartDate := metav1.NewTime(time.Time{})
+	lastRestartableInitContainerRestartDate := metav1.NewTime(time.Time{})
```
obs-gh-mattcotter, line 86: A lot of these new variables don't seem to be used for anything; is this coming in a follow up?

components/processors/observek8sattributesprocessor/podactions_test.go
```diff
@@ -12,133 +12,133 @@ func TestPodActions(t *testing.T) {
...
-			),
-			expectedResults: []queryWithResult{
-				// Conditions must be a map with 5 elements
-				{
-					path:      "observe_transform.facets.conditions | length(@)",
-					expResult: float64(5),
-				},
 			},
 		},
+		// { // Tests that we don't override/drop other facets computed in OTTL
```
obs-gh-mattcotter, line 22: Can you replace these deleted tests with new ones?

components/processors/observek8sattributesprocessor/podactions.go
```diff
@@ -43,18 +75,46 @@ func NewPodStatusAction() PodStatusAction {
 
 // Generates the Pod "status" facet.
 func (PodStatusAction) ComputeAttributes(pod v1.Pod) (attributes, error) {
-	// based on https://github.com/kubernetes/kubernetes/blob/0d3b859af81e6a5f869a7766c8d45afd1c600b04/pkg/printers/internalversion/printers.go#L901
-	reason := string(pod.Status.Phase)
+	// based on https://github.com/kubernetes/kubernetes/blob/cd3b5c57668a0a6e32057ef82dfab40e9b0bec5b/pkg/printers/internalversion/printers.go#L881
+	restarts := 0
+	restartableInitContainerRestarts := 0
+	totalContainers := len(pod.Spec.Containers)
+	readyContainers := 0
```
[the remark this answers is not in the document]
obs-gh-alexlew: so here's the logic in datadog agent https://github.com/DataDog/datadog-agent/blob/7d47d25899c025c8b3e5ba91e488fa0f97f7a275/pkg/collector/corechecks/cluster/orchestrator/transformers/k8s/pod.go#L220 they also only use status but kept the rest of the logic so that they can just copy paste it over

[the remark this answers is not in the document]
obs-gh-alexlew: think the argument for taking it out is that its unnecessary computation / more efficient which i agree with but its a slight preference vs. way better imo

components/processors/observek8sattributesprocessor/podactions.go
obs-gh-mattcotter: Bumping this one

components/processors/observek8sattributesprocessor/podactions.go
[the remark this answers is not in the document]
obs-gh-mattcotter: Interesting. I'm happy with where you landed!

If we need to change it again, I think it could be more clear to have a helper that returns all values and is very explicitly full copy/paste. Then we could call that and extra only what we want

obs-gh-mattcotter (approved):

PR #215, opened 2025-06-05
feat: add default multiline autodetection

obs-gh-alexlew:
### Description

### Checklist

### DIFF
--- a/packaging/docker/observe-agent/connections/host_monitoring/logs.yaml.tmpl
+++ b/packaging/docker/observe-agent/connections/host_monitoring/logs.yaml.tmpl
@@ -27,7 +27,7 @@ receivers:
       - id: multiline-recombine
         type: recombine
         combine_field: body
-        # Regex is just 4 different pattern's OR'd together to match the 4 timestamp formats `2021-03-28 13:45:30`, `2023-03-28T14:33:53.743350Z`, `Jun 14 15:16:01`, `2024/05/16 19:46:15`
+        # Regex is just 3 different pattern's OR'd together to match the 4 timestamp formats `2021-03-28 13:45:30`, `2023-03-28T14:33:53.743350Z`, `Jun 14 15:16:01`, `2024/05/16 19:46:15`
         is_first_entry: body matches "^(\\d{4}[-\\/]\\d{2}[-\\/]\\d{2} \\d{2}:\\d{2}:\\d{2}|\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(\\.\\d+)?Z?|[A-Za-z]+ \\d{1,2} \\d{2}:\\d{2}:\\d{2})"
       {{- end }}
 
--- a/packaging/docker/observe-agent/connections/host_monitoring/logs.yaml.tmpl
+++ b/packaging/docker/observe-agent/connections/host_monitoring/logs.yaml.tmpl
@@ -23,6 +23,12 @@ receivers:
     operators:
       - type: filter
         expr: 'body matches "otel-contrib"'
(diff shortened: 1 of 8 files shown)


packaging/docker/observe-agent/connections/host_monitoring/logs.yaml.tmpl
```diff
@@ -23,6 +23,12 @@ receivers:
     operators:
       - type: filter
         expr: 'body matches "otel-contrib"'
+      {{- if .HostMonitoring.Logs.AutoMultilineDetection }}
+      - id: multiline-recombine
+        type: recombine
+        combine_field: body
+        is_first_entry: body matches "^(\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}|\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(\\.\\d+)?Z?|[A-Za-z]+ \\d{1,2} \\d{2}:\\d{2}:\\d{2}|\\d{4}\\/\\d{2}\\/\\d{2} \\d{2}:\\d{2}:\\d{2})"
```
obs-gh-mattcotter, line 30: A few comments on the regex:

- can we document what this is and where it came from?
- is `\\/` right for forward slash? I'm surprised it needs to be escaped at all.
- is the space vs `T` important? Can we combine those?
- is `-` vs `/` important for dates? can we combine them?
  - ex for my last two combination requests: `\d{4}[/-]\d{2}[/-]\d{2}[T ]\d{2}:\d{2}:\d{2}(\.\d+)?Z?` and then it could be OR'ed with `[A-Za-z]+ \d{1,2} \d{2}:\d{2}:\d{2}` like you have it.

packaging/docker/observe-agent/connections/host_monitoring/logs.yaml.tmpl
```diff
@@ -23,6 +23,12 @@ receivers:
     operators:
       - type: filter
         expr: 'body matches "otel-contrib"'
+      {{- if .HostMonitoring.Logs.AutoMultilineDetection }}
+      - id: multiline-recombine
```
obs-gh-mattcotter, line 27: Does this operator differ from the top level `multiline` config option?
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.124.0/receiver/filelogreceiver#example---multiline-logs-parsing
If they work the same, I think it's better to use the `multiline` option since it can be overridden by users more easily than a field in an array.

obs-gh-alexlew: so i think specifically the order of operations between the multiline top level config and then the operators pipelines means that we need to define this as another operator after the container operator for kubernetes specifically. I dont actually know if its necessary in this case but figured it might be best to do it the same way everywhere so that the behavior is most likely to be similar.

packaging/docker/observe-agent/connections/host_monitoring/logs.yaml.tmpl
obs-gh-alexlew: these are just based on the 4 timestamps that datadog supports https://docs.datadoghq.com/agent/logs/auto_multiline_detection/?tab=configurationfile#enable-multi-line-aggregation-per-integration and they're just OR'd together

obs-gh-alexlew: i've merged the regexs for slashes versus dashes, the other ones I think its clearer to leave them separate since they're less close

obs-gh-mattcotter (approved):

PR #216, opened 2025-06-05
feat: support otelcol flags (config, set, feature-gates) in all sub commands

obs-gh-mattcotter:
### Description

Support otelcol flags (`config`, `set`, and `feature-gates`) in all sub commands. Substitutes of the listed otel flags are now configured at the root command level on the observe-agent and are applied to all subcommands.


internal/commands/config/config_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/observeinc/observe-agent/internal/commands/util/logger"
+	"github.com/observeinc/observe-agent/internal/root"
+	"github.com/observeinc/observe-agent/observecol"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+)
+
+// TODO enable this test once the agent bundled configs don't depend on the host filesystem having the templates.
+func XTest_RenderOtelConfig(t *testing.T) {
```
obs-gh-alexlew, line 24: maybe we should have some sort of unit test mode that looks for the templates somewhere relative to current path vs. somewhere specific on the FS. im fine with just looking at linux and not adding yet another copy of everything

observecol/flags.go
```diff
@@ -0,0 +1,24 @@
...
+	setFlagDescription = "Set arbitrary OpenTelemetry component config properties. The component has to be defined in " +
+		"the bundled or provided OpenTelemetry config files and this flag has a higher precedence. " +
+		"Array config properties are overridden and maps are joined. Example --set=processors.batch.timeout=2s"
+)
+
+var otelConfigs []string
+var otelSets []string
+
+func AddConfigFlags(flags *pflag.FlagSet) {
+	flags.StringSliceVar(&otelConfigs, configFlag, []string{}, configFlagDescription)
```
obs-gh-alexlew, line 22: what does this actually do?

obs-gh-alexlew (approved):

PR #225, opened 2025-06-12
feat: enable filelog receiver in windows if include is specified

obs-gh-mattcotter:
### Description

OB-46154 enable filelog receiver in windows if `host_monitoring.logs.include` is specified.

Depends on #223

### DIFF
--- a/packaging/windows/connections/host_monitoring/logs.yaml.tmpl
+++ b/packaging/windows/connections/host_monitoring/logs.yaml.tmpl
@@ -14,6 +14,35 @@ receivers:
     retry_on_failure:
       enabled: true
 
+  {{- if .HostMonitoring.Logs.Include }}
+  filelog/host_monitoring:
+    include:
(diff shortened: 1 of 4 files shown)


packaging/windows/connections/host_monitoring/logs.yaml.tmpl
```diff
@@ -14,6 +14,35 @@ receivers:
     retry_on_failure:
       enabled: true
 
+  {{- if .HostMonitoring.Logs.Include }}
+  filelog/host_monitoring:
+    include:
+      {{- range .HostMonitoring.Logs.Include }}
```
obs-gh-alexlew, line 20: what are the default values for these?

obs-gh-mattcotter: By default the monitoring only checks `windowseventlog`. If `HostMonitoring.Logs.Include` isn't set, this receiver won't be included.

PR #227, opened 2025-06-17
feat: add jsonschema for config file and fix readme

obs-gh-alexlew:
### Description

### Checklist


observe-agent.schema.json
```diff
@@ -0,0 +1,193 @@
+{
```
obs-gh-mattcotter, line 1: Good idea! Is this / can this be autogenerated?

observe-agent.schema.json
```diff
@@ -0,0 +1,193 @@
...
+          "type": "boolean",
+          "description": "Enable or disable telemetry forwarding."
+        },
+        "metrics": {
+          "type": "object",
+          "description": "Forwarding options for metrics.",
+          "properties": {
+            "output_format": {
+              "type": "string",
+              "description": "Format used for outputting metrics (e.g., otel)."
```
obs-gh-mattcotter, line 111: Can this use a pattern to validate the string? https://json-schema.org/draft/2020-12/json-schema-validation#name-pattern
ex:
```suggestion
              "description": "Format used for outputting metrics (e.g., otel).",
              "pattern": "otel|prometheus"
```
Not sure if I got the syntax right, but something like that.

obs-gh-mattcotter (approved):

observe-agent.schema.json
obs-gh-mattcotter: That would be awesome!

internal/config/configschema.go
```diff
@@ -79,7 +79,7 @@ type InternalTelemetryConfig struct {
 }
 
 type AgentConfig struct {
-	Token                  string                  `yaml:"token" mapstructure:"token"`
+	Token                  string                  `yaml:"token" mapstructure:"token" jsonschema:"required"`
 	ObserveURL             string                  `yaml:"observe_url" mapstructure:"observe_url"`
```
obs-gh-mattcotter, line 83: This is required too

observe-agent.schema.json
obs-gh-mattcotter: One other thing looking forward with this file; can we include it in our release? (like this: https://github.com/observeinc/observe-agent/blob/main/.goreleaser.yaml#L192)
Then we can have people point at the latest released version and not worry about making changes on the main branch.

obs-gh-alexlew: not sure what you mean by this, currently you just need to run the makefile to generate the file when you make schema changes. would be nice to automate this and commit the changes on its own later

obs-gh-alexlew: ah i think i understand what you mean, we should bundle this file with the actual installation releases. I think we can do that later, there's not actually a runtime use case for this file rn. if we come up with one we can add it then.

obs-gh-mattcotter: I was thinking we could put
```yaml
# yaml-language-server: $schema=https://github.com/observeinc/observe-agent/releases/latest/download/observe-agent.schema.json
```
in our agent-config.yaml (ie emit this with the `init-config`, add it to docs, etc).
And in this case, it's better to point to `releases/latests` instead of `refs/heads/main`

PR #229, opened 2025-06-25
refactor: remove all receiver references to bundled config host files

obs-gh-mattcotter:
### Description

OB-42646 Remove all receiver references to bundled config fragments. These files are going away in favor of adding them to the agent binary.

### DIFF
--- a/internal/commands/config/test/snap1-docker-output.yaml
+++ b/internal/commands/config/test/snap1-docker-output.yaml
@@ -131,14 +131,6 @@ processors:
                 - truncate_all(attributes, 4095)
                 - truncate_all(resource.attributes, 4095)
 receivers:
-    filelog/agent-config:
-        include:
-            - /etc/observe-agent/connections/common/base.yaml.tmpl
-        include_file_path: true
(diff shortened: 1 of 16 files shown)


obs-gh-mattcotter:
> I think this does drive the host monitoring app's resource dataset which customers have complained about breaking before

Good point! I put this receiver back and directed it at the agent config. I tested this locally with the otel app and it does work; it successfully tracks the host as a resource.

internal/commands/config/test/snap1-docker-output.yaml
```diff
@@ -160,7 +152,7 @@ receivers:
         storage: file_storage
     filestats/agent:
         collection_interval: 5m
-        include: /etc/observe-agent/connections/common/base.yaml.tmpl
+        include: /etc/observe-agent/observe-agent.yaml
```
obs-gh-alexlew, line 155: this should always exist right? unless the user for some reason manually deletes it after install

obs-gh-alexlew (approved):

PR #234, opened 2025-07-16
feat: add the yaml schema comment to our default agent configs and init-config

obs-gh-mattcotter:
### Description

IDEs check this to enforce yaml schema. I think that it is safe to release the schema for the first time in the same release that we set this comment, since the comment will not be in use until the schema is released.


packaging/linux/config/observe-agent.yaml
```diff
@@ -1,3 +1,5 @@
+# yaml-language-server: $schema=https://github.com/observeinc/observe-agent/releases/latest/download/observe-agent.schema.json
```
obs-gh-alexlew, line 1: is it okay for this to point to latest? don't we want this to point to a specific version until we draft the next version?

PR #237, opened 2025-07-24
feat: add an option to generate RED metrics from spans

obs-gh-mattcotter:
This is similar to https://github.com/observeinc/helm-charts/pull/394 but with the k8s logic replaced by host resource detection.

### DIFF
--- a/internal/connections/bundledconfig/docker/common/base.yaml.tmpl
+++ b/internal/connections/bundledconfig/docker/common/base.yaml.tmpl
@@ -116,4 +116,9 @@ service:
     metrics/count-nop-out:
        receivers: [count]
        exporters: [nop]
-  extensions: [health_check, file_storage, cgroupruntime]
+  extensions:
+    {{- if .HealthCheck.Enabled }}
+    - health_check
+    {{- end }}
+    - file_storage
+    - cgroupruntime
--- a/internal/connections/bundledconfig/shared/common/forward.yaml.tmpl
+++ b/internal/connections/bundledconfig/shared/common/forward.yaml.tmpl
@@ -6,6 +6,25 @@ receivers:
       http:
         endpoint: localhost:4318
 
+processors:
+  transform/add_span_status_code:
+    error_mode: ignore
+    trace_statements:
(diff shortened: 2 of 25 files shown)


internal/connections/bundledconfig/docker/common/base.yaml.tmpl
```diff
@@ -116,4 +116,9 @@ service:
     metrics/count-nop-out:
        receivers: [count]
        exporters: [nop]
-  extensions: [health_check, file_storage, cgroupruntime]
+  extensions:
+    {{- if .HealthCheck.Enabled }}
+    - health_check
+    {{- end }}
```
obs-gh-mattcotter, line 122: The collector errors if the health check extension is defined but not configured.

internal/connections/bundledconfig/shared/common/forward.yaml.tmpl
```diff
@@ -6,6 +6,25 @@ receivers:
       http:
         endpoint: localhost:4318
 
+processors:
+  transform/add_span_status_code:
```
obs-gh-mattcotter, line 10: Both of these are features added in the helm chart to facilitate Service Explorer. I'm adding them in the same PR because they affect RED metrics generation.

scripts/generate_jsonschema.go
```diff
@@ -12,7 +13,14 @@ import (
 
 func main() {
 	r := new(jsonschema.Reflector)
-	r.KeyNamer = strcase.SnakeCase // from package github.com/stoewer/go-strcase
+	r.KeyNamer = func(name string) string {
```
obs-gh-mattcotter, line 16: I needed to customize this to allow the key `RED_metrics`

obs-gh-alexlew:
does this potentially remove any attributes that were present before?

obs-gh-alexlew (approved):

obs-gh-mattcotter:
> does this potentially remove any attributes that were present before?

Nope! The only change to old data is the new filter processor to drop spans with a duration longer than the configured max span time, and to compute `status_code` and add it as an attribute to spans (but only if they don't already have one). Everything else is for the new RED metrics pipelines.

PR #240, opened 2025-08-11
fix: use env as default config resolve schema

obs-gh-mattcotter:
### Description

Use "env" as the default config resolver schema so `${TOKEN}` in passed in config will read from an environment variable (just like `${env:TOKEN}`).


obs-gh-mattcotter:
Fwiw I couldn't find the proper OTel docs on this. Their [config doc's environment variables section](https://opentelemetry.io/docs/collector/configuration/#environment-variables) only ever uses `env:`, so if anyone knows if/where this is mentioned in OTel docs, please lmk and I'll link it.

obs-gh-alexlew (approved):

PR #256, opened 2025-09-02
Add restart on install to linux postinstall/postremove scripts

obs-gh-alexlew:
### Description

### Checklist

### DIFF
--- a/packaging/linux/scripts/postinstall.sh
+++ b/packaging/linux/scripts/postinstall.sh
@@ -12,14 +12,22 @@ if command -v systemctl >/dev/null 2>&1; then
     # We'll check if the service exists and is active to determine if this is an upgrade
 
     if systemctl list-unit-files observe-agent.service >/dev/null 2>&1; then
-        # Service file exists, check if it's currently active
+        # Service file exists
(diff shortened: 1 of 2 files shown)


packaging/linux/scripts/postinstall.sh
```diff
@@ -1,3 +1,40 @@
 #!/bin/sh
 
 sudo setcap 'cap_dac_read_search=ep' /usr/bin/observe-agent
+
+# Check if systemd is available and handle service restart for upgrades
+if command -v systemctl >/dev/null 2>&1; then
```
obs-gh-mattcotter, line 6: We should unify this check with the check here
https://github.com/observeinc/observe-agent/blob/main/scripts/install_linux.sh#L131-L132

Can you get that in this change?

obs-gh-alexlew: By unify, do you just mean standardize on the exact check we run?

obs-gh-mattcotter: Yes, I think we should pick one or the other and make all instances of the "are we using systemd" the same check

PR #267, opened 2025-09-18
fix: update postinstall script syntax

obs-gh-alexlew:
### Description

### Checklist


obs-gh-alexlew:
I copied the check from the other script but it turns out these are `/bin/sh` scripts whereas the other script is `/bin/bash` and for compatibility the pre/post install scripts are better as `/sh`. I've updated the syntax to be correct for that

PR #307, opened 2026-02-12
feat: upgrade observe-agent health check extension to health check v2

obs-gh-ruoyuchen:
### Description

Upgrade observe-agent health check extension to health check v2


obs-gh-mattcotter (commented):
Looks great! A few comments on top of my two inline:
1. Can you update the table in the README to reflect the new extension?
2. Can you enable this test locally and make sure it passes: https://github.com/observeinc/observe-agent/blob/main/internal/commands/config/config_test.go#L102

PR #321, opened 2026-04-01
chore: add new skill to handle observe-agent releases

obs-gh-mattcotter:
This was created from our internal process documentation and a session with the agent actually doing a release.

### DIFF
--- a/.agents/skills/release-agent/SKILL.md
+++ b/.agents/skills/release-agent/SKILL.md
@@ -1,0 +1,132 @@
+---
+name: release-agent
+description: >-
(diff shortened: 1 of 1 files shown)


obs-gh-virjramakrishnan (approved):
Looks great! In your experience, was the agent able to navigate to O2 and get data with this file?

obs-gh-mattcotter:
> Looks great! In your experience, was the agent able to navigate to O2 and get data with this file?

Originally I hadn't tried that step since I created the skill after getting that far on my own, but I ran through it again (just this verification step) and was able to get it to work with a slight modification. Great catch, thanks!

PR #325, opened 2026-04-17
chore: add AI agent for code review

obs-gh-abhishekrao:
### Description

- Introduces a code-review subagent. 
- Adds a CI workflow using Cursor to review pull requests

### Checklist

### DIFF
--- a/.github/workflows/code-review.yaml
+++ b/.github/workflows/code-review.yaml
@@ -1,0 +1,247 @@
+name: Cursor Code Review
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  workflow_dispatch:
+    inputs:
+      model:
+        description: "Model for Cursor agent CLI"
+        required: false
+        default: "gpt-5.3-codex"
+        type: string
+
+permissions:
+  contents: read
+  pull-requests: write
+
+concurrency:
+  group: cursor-code-review-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  review:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Preflight
+        id: preflight
+        env:
+          # Check secret availability via env mapping — you can't reference
+          # `secrets.*` directly in a step `if:` condition.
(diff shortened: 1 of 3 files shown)


.github/workflows/code-review.yaml
```diff
@@ -0,0 +1,247 @@
...
+          curl https://cursor.com/install -fsS | bash
+          echo "$HOME/.cursor/bin" >> "$GITHUB_PATH"
+
+      - name: Run code-reviewer
+        if: steps.preflight.outputs.skip != 'true'
+        env:
+          CURSOR_API_KEY: ${{ secrets.CURSOR_API_KEY }}
+        run: |
+          read -r -d '' PROMPT << 'EOF' || true
+          Review the unified diff in ./diff.patch (summary in ./diff.stat) for repository ${{ github.repository }}.
```
obs-gh-mattcotter, line 72: Seems like this prompt duplicates the subagent md file. Do we need to repeat all of this?

obs-gh-abhishekrao: I don't think it's duplicated. Cursor's CLI keeps throwing model errors when invoking the agent directly using `agent -p "/code-reviewer <rest of prompt>`, so I had to directly invoke it as a main agent, thus needing a "prompt".


Effectively what it is:
- Invokes Cursor agent, tells it to read the `code-reviewer` agent definition and follow what's written there
- Additional instructions to adjust formatting according to how we want to see PR comments.
It may seem like it's duplicated but it's actually not. The core parts are still going to use the agent definition.

obs-gh-abhishekrao: I'll also check if there's better logging functionality available in the agent command, so we can better track agent behaviour and explain it more concretely.

obs-gh-virjramakrishnan: > Additional instructions to adjust formatting according to how we want to see PR comments.

Any reason why we can't include these in the agent description? (I mean lines 78-130)

obs-gh-abhishekrao: @obs-gh-virjramakrishnan I'll work on that. My intention was - PR comments may need to be more condensed down so a human reviewer sees the important bits, while local agent runs are detailed enough for a contributor to action or work with (or other agents to iterate commits locally). That said, the way it's currently written looks like it's mostly just a duplication of what's in the agent definition, so I'll refine it to fit immediate needs.

PR #326, opened 2026-04-22
feat(deps): upgrade otelcol to v0.151.0 and go to 1.25.9

obs-gh-abhishekrao:
### Description

- Resolve CVEs discovered on `main` - https://github.com/observeinc/observe-agent/actions/runs/24789374709

### Checklist


.orca-security.yaml
```diff
@@ -0,0 +1,4 @@
+global:
+  excluded_paths:
+    - "./vendor/**/package-lock.json"
```
obs-gh-abhishekrao, line 3: Strangely the CI excludes already has `vendor` but I had to add this. Keeping it scoped to package / package-lock for now since those are unrelated to Go.

obs-gh-mattcotter (commented):
The change looks good to me. Any breaking changes upstream we should look out for?

obs-gh-abhishekrao:
> The change looks good to me. Any breaking changes upstream we should look out for?

There are quite a few breaking changes. In terms of what affects us, there's one that needs a solution in this repo and one that _may_ need addressing in the helm chart.
> 1. v0.150.0 (Contrib):
pkg/ottl (affects transform/filter processors): Setters now validate value types and return errors instead of silently failing. Existing OTTL statements with type mismatches will start erroring.
 > 2. v0.149.0 (Core):
service: Removed service_name, service_instance_id, service_version as constant labels on every internal metric datapoint. These are now only in target_info. Users scraping the collector's internal metrics endpoint will need to use target_info joins instead.

Edit: Clarified exact changes. Never mind, we're not affected by these either. The OTTL is a close one but our helm charts are already protected the way OTTL is written.

obs-gh-mattcotter (approved):

obs-gh-abhishekrao:
Windows tests failed because of this - https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/f52b9a96d9ea179e25b302e98da03e48779c0caa

We can just use the new naming and avoid waiting for a fix.
Edit: Existing configs would break, so we can't do this :-(

observecol/compat.go
```diff
@@ -0,0 +1,67 @@
...
+		wrapped := &deprecatedAliasCompat{Factory: f}
+		factories.Receivers[typ] = wrapped
+		factories.Receivers[alias] = wrapped
+		if mod, ok := factories.ReceiverModules[typ]; ok {
+			factories.ReceiverModules[alias] = mod
+		}
+	}
+}
+
+func componentsWithCompat() (otelcol.Factories, error) {
```
obs-gh-mattcotter: Can you add a TODO to remove this once the alias is added back upstream?

obs-gh-mattcotter (approved):
lgtm, let's make sure to do extra testing on this before releasing though

observecol/compat.go
obs-gh-abhishekrao: Since the upstream fix is now available, this is no longer needed and I have completely removed it.

obs-gh-abhishekrao:
> Windows tests failed because of this - [open-telemetry/opentelemetry-collector-contrib@f52b9a9](https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/f52b9a96d9ea179e25b302e98da03e48779c0caa)
> 
> We can just use the new naming and avoid waiting for a fix. Edit: Existing configs would break, so we can't do this :-(

This is now fixed upstream in 0.151.0.

obs-gh-mattcotter (approved):

obs-gh-mattcotter (approved):

PR #327, opened 2026-04-22
fix: use a patched version of k8sattributesprocessor to address memory issues

obs-gh-mattcotter:
### Description

The k8sattributesprocessor upstream has an issue where the memory will increase unboundedly with every new pod seen. This PR applies the [fix here](https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/f608be1add846bd6f04199031e103996b39594b2) (to disable the metric causing the issue) to the processor version in the observe-agent.

The diff of the vendored module is small and matches the upstream commit. I ran this locally and verified the metric is no longer generated.


obs-gh-abhishekrao (approved):

obs-gh-mattcotter:
Closing because we now have the upstream patch

PR #328, opened 2026-04-24
fix: default docker image to bind 0.0.0.0 instead of localhost

sfc-gh-zeningchen:
## Summary
- The container image defaulted all listener endpoints (OTLP gRPC/HTTP, health check, internal telemetry) to `localhost`, making them unreachable from outside the container.
- Override these addresses to `0.0.0.0` in the docker-packaged `observe-agent.yaml` so other containers, Kubernetes probes, and Prometheus scrapers can connect out of the box.
- No impact on bare-metal Linux, macOS, or Windows installs — their code-level defaults remain `localhost`.
- Add an integration test that spins up a second container on the same Docker network and verifies all three endpoints are reachable from outside the agent container.

## Test plan

- [x] Local Docker test: verified all 3 endpoints reachable with `0.0.0.0`, and unreachable with `localhost` (see results below)
- [x] CI `test_docker_networking` integration test validates health check (13133), internal telemetry (8888), and OTLP HTTP (4318) from a separate container
- [x] Existing snapshot tests unaffected (config-only change, not referenced by tests)

## Local test results

**Image:** `observeinc/observe-agent:latest` (v2.15.0)
**Method:** Start agent in a container, curl endpoints from a separate container on the same Docker network.

### With `0.0.0.0` bindings (the fix)

Config overrides:
```yaml
health_check:
  endpoint: 0.0.0.0:13133
forwarding:
  endpoints:
    http: 0.0.0.0:4318
    grpc: 0.0.0.0:4317
internal_telemetry:
  metrics:
    host: 0.0.0.0
```

Agent logs confirm listeners on all interfaces:
```
Starting GRPC server  {"endpoint": "[::]:4317"}
Starting HTTP server  {"endpoint": "[::]:4318"}
Health Check state change  {"status": "ready"}
Everything is ready. Begin running and processing data.
```

**Test 1: Health Check (port 13133) — PASS**
```
$ docker run --rm --network test-net curlimages/curl curl -sf http://agent:13133/status
{"status":"Server available","upSince":"2026-04-27T04:18:05.415025606Z","uptime":"13.774787092s"}
Exit code: 0
```

**Test 2: Internal Telemetry (port 8888) — PASS**
```
$ docker run --rm --network test-net curlimages/curl curl -sf http://agent:8888/metrics
# HELP otelcol_deltatocumulative_streams_limit upper limit of tracked streams
# TYPE otelcol_deltatocumulative_streams_limit gauge
[...]

PR #329, opened 2026-04-27
fix: default Docker listener endpoints to 0.0.0.0 via Viper defaults

sfc-gh-zeningchen:
## Summary
- Follow-up to #328 (79238b8). The previous fix updated the packaged `observe-agent.yaml` to bind listeners to `0.0.0.0` in Docker, but customers who supply their own config file still fell back to the hardcoded `localhost` defaults, making the agent unreachable from outside the container.
- Add `setDockerDefaults()` in `root.go` that overrides Viper defaults for forwarding gRPC/HTTP, health check, and internal telemetry endpoints to `0.0.0.0` when `--config-mode=docker`. This ensures the agent binds to all interfaces even when a custom `observe-agent.yaml` omits these settings.
- Remove the now-redundant `0.0.0.0` overrides from the packaged `packaging/docker/observe-agent/observe-agent.yaml` since the Viper defaults handle this.
- Customer-provided values in the config file still take precedence over these defaults.

## Test plan
- [x] Updated snapshot tests to set `config-mode` flag for each package type, matching production behavior
- [x] Docker snapshots (`snap0-docker-output.yaml`, `snap3-docker-output.yaml`) updated to expect `0.0.0.0` endpoints
- [x] `snap1-docker-output.yaml` (full config with explicit endpoints) unaffected — confirms customer overrides still work
- [x] All Linux/macOS/Windows snapshots unchanged — confirms no impact on non-Docker modes
- [x] All 14 snapshot tests pass

## Local Docker test results

**Setup:** Built the Docker image with the reverted `observe-agent.yaml` (no `0.0.0.0` overrides in the config file). Started the agent with `--config-mode docker` and tested from a separate container on the same Docker network.

Agent logs confirm listeners on all interfaces (from `setDockerDefaults()` Viper defaults only):
```
Starting GRPC server  {"endpoint": "[::]:4317"}
Starting HTTP server  {"endpoint": "[::]:4318"}
Health Check state change  {"status": "ready"}
Everything is ready. Begin running and processing data.
```

**Test 1: Health Check (port 13133) — PASS**
```
$ docker run --rm --network test-net curlimages/curl curl -sf http://agent-test:13133/status
{"status":"Server available","upSince":"2026-04-27T20:03:41.629448741Z","uptime":"2m46.892990843s"}
Exit code: 0
```

**Test 2: Internal Telemetry (port 8888) — PASS**
```
$ docker run --rm --network test-net curlimages/curl curl -sf http://agent-test:8888/metrics
# HELP http_client_request_body_size_bytes Size of HTTP client request bodies.
# TYPE http_client_request_body_size_bytes histogram
... (200+ lines of Prometheus metrics)
Exit code: 0
```

**Test 3: OTLP HTTP Receiver (port 4318) — PASS**
```
$ docker run --rm --network test-net curlimages/curl \
    curl -sf -X POST http://agent-test:4318/v1/traces \
    -H 'Content-Type: application/json' -d '{"resourceSpans":[]}'
{"partialSuccess":{}}
Exit code: 0
```

| Endpoint | Port | Result |
|---|---|---|
| Health check | 13133 | 200 OK |
| Internal telemetry | 8888 | 200 OK |
| OTLP HTTP receiver | 4318 | 200 OK |

### DIFF
--- a/internal/root/root.go
+++ b/internal/root/root.go
@@ -78,6 +78,13 @@ func setConfigMode() {
 
 }
 
+func getConfigMode() string {
+	if configMode != "" {
+		return strings.ToLower(configMode)
+	}
+	return bundledconfig.ConfigEnvironment
+}
+
 // InitConfig reads in config file and ENV variables if set.
 func InitConfig() {
 	setConfigMode()
@@ -99,7 +106,7 @@ func InitConfig() {
 	// viper.SetEnvPrefix("OBSERVE")
 	viper.AutomaticEnv() // read in environment variables that match
 
-	config.SetViperDefaults(viper.GetViper(), "::")
+	config.SetViperDefaults(viper.GetViper(), "::", getConfigMode())
 
 	// If a config file is found, read it in.
 	if err := viper.ReadInConfig(); err != nil {
--- a/internal/commands/config/config_test.go
+++ b/internal/commands/config/config_test.go
@@ -203,6 +203,8 @@ func setupConfig(t *testing.T, test snapshotTest) {
 	})
 
 	root.CfgFile = filepath.Join(curPath, test.agentConfigPath)
+	root.RootCmd.PersistentFlags().Set("config-mode", string(test.packageType))
+	t.Cleanup(func() { root.RootCmd.PersistentFlags().Set("config-mode", "") })
 	root.InitConfig()
 
 	setEnvVars(t, test.packageType)
--- a/internal/commands/config/test/snap0-docker-output.yaml
+++ b/internal/commands/config/test/snap0-docker-output.yaml
@@ -54,7 +54,7 @@ extensions:
     file_storage:
         directory: /var/lib/observe-agent/filestorage
     health_check:
-        endpoint: localhost:13133
+        endpoint: 0.0.0.0:13133
         path: /status
 processors:
     batch:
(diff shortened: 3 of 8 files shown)


internal/root/root.go
```diff
@@ -78,6 +78,23 @@ func setConfigMode() {
 
 }
 
+// setDockerDefaults overrides listener endpoint defaults to 0.0.0.0 so the
+// agent is reachable from outside the container even when a customer-provided
+// observe-agent.yaml omits these settings.
+func setDockerDefaults() {
```
obs-gh-mattcotter: I think it might be cleaner to add this to `SetViperDefaults` here: https://github.com/observeinc/observe-agent/blob/main/internal/config/configschema.go#L153

You might need to add the configMode as an argument to that method.

obs-gh-mattcotter (commented):
One comment about where to do this. The test changes look good to me!

internal/root/root.go
sfc-gh-zeningchen: that's a good point! I've updated the code accordingly.

obs-gh-mattcotter (approved):
Nice! Lgtm, thanks

obs-gh-abhishekrao (approved):

PR #330, opened 2026-04-29
chore: improve goreleaser builds for integration tests

obs-gh-abhishekrao:
### Description

Removes unused targets from goreleaser in integration tests. This should significantly speed up the build. A separate goreleaser config is introduced for ci purposes for easier control.

### Checklist

### DIFF
--- a/.github/workflows/tests-integration.yaml
+++ b/.github/workflows/tests-integration.yaml
@@ -4,9 +4,9 @@ on:
   workflow_dispatch:
     inputs:
       old_version:
-        description: 'Old version of observe-agent to test upgrade from (e.g., v2.5.0)'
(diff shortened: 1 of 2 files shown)


.goreleaser.tests.yaml
obs-gh-mattcotter, line 1: How can we ensure that this stays in sync with our main goreleaser file? Can we skip some steps in the main one if we pass a test flag? Or generate this file from the other one with a python script? It would be best to have a single source of truth for our goreleaser build, otherwise we risk not testing the binary we would later release.

obs-gh-abhishekrao: Updated the workflow to use a dynamically generated config using `yq`.

obs-gh-mattcotter (approved):
Awesome!

PR #338, opened 2026-05-07
fix: quote attribute/resource_attribute values in bundled-config templates

sfc-gh-zeningchen:
## Summary

The bundled-config templates that render the operator's `attributes:` and `resource_attributes:` blocks into the otelcol `attributes`/`resource` processor configs emitted **unquoted** YAML scalars:

```
- key: {{ $key }}
  value: {{ $value }}            # bare scalar — string typing lost
```

When otelcol then loaded the rendered config, its YAML parser interpreted any value that happened to look like a typed scalar as that type. `resourceprocessor.Action.Value` is `interface{}`, so the typed value was stamped onto every span, metric, and log going through the affected pipelines.

The fix routes each value through the existing `toYaml` template helper (`internal/utils/templatefuncs.go`), which marshals through `goccy/go-yaml` and returns a canonical YAML scalar — guaranteed to be a valid scalar by construction, with proper handling of values that contain quotes, newlines, or other characters that need escaping.

## Concrete cases observed (OB-60193)

| Operator setting | Without fix → on signals | With fix → on signals |
|---|---|---|
| `service.version: "060520262036"` | `int64 6,530,622,494` (octal) | `"060520262036"` ✓ |
| `service.version: "060520261956"` | `int64 60,520,261,956` (decimal, leading 0 stripped) | `"060520261956"` ✓ |
| `team.name: "no"` | `bool false` (YAML 1.1 "Norway problem") | `"no"` ✓ |
| `service.namespace: "1.0"` | `float64 1` | `"1.0"` ✓ |
| Values with `:` `#` `[` `]` `{` `}` `,` | parse breaks / shape changes | preserved ✓ |

## Affected pipelines

Anything using `resource/observe_global_resource_attributes` or `attributes/observe_global_attributes`:

- traces / metrics / logs forward (`forward.yaml.tmpl`)
- host & process metrics, host logs (Linux/Docker/Windows variants)
- R.E.D metrics, self-monitoring, fleet heartbeat

So the same misrendered attribute lands wrong on **every** signal — not just one dataset.

## Origin

Discovered during validation of 4 Java Spring Boot messaging apps where `sample-app-validator` sets a digit-only timestamp `service.version`. The empty Tracing/Deployment dataset led upstream to this template. See OB-60193 for the full investigation, including the Python SDK reproduction that isolated the bug to the agent (Observe backend correctly preserves strings that arrive intact on the wire).

## Test plan

- [x] `go test ./internal/... ./components/...` — all packages PASS
- [x] `Test_RenderOtelConfig` — 14/14 snapshot tests PASS
- [x] Extended `snap1-full-agent-config.yaml` with regression fixtures covering each coercion class still active under YAML 1.2 (otelcol's `confmap` loader uses `gopkg.in/yaml.v3`, which dropped the YAML 1.1 `yes/no/on/off` boolean aliases — so the Norway-problem case from the bug report no longer manifests through this loader and was intentionally omitted as a test case):

  | Input | Without fix | With fix |
  |---|---|---|
  | `bool-attr: "true"` | `value: true` (bool) | `value: "true"` ✓ |
  | `digits-attr: "12345"` | `value: 12345` (int64) | `value: "12345"` ✓ |
  | `service.namespace: "1.0"` | `value: 1` (float→int) | `value: "1.0"` ✓ |
  | `service.version: "060520262300"` | `value: 6530622656` (octal int64) | `value: "060520262300"` ✓ |

- [x] Verified each of the four added cases is regression-catching: with the inputs in place, reverting just the template change produces a diff on every line above.
- [x] `observe-agent config` against `/etc/observe-agent/observe-agent.yaml` with digit-only `service.version` — rendered output now contains `value: "060520262300"` (verified manually before this PR).


internal/connections/bundledconfig/shared/common/attributes.yaml.tmpl
```diff
@@ -4,7 +4,7 @@ processors:
     actions:
     {{- range $key, $value := .Attributes }}
       - key: {{ $key }}
-        value: {{ $value }}
+        value: {{ printf "%q" $value }}
```
obs-gh-mattcotter: Nice find! Can this use the `toYaml` template function instead of printf? I don't know if %q is guaranteed to match the yaml format.

sfc-gh-zeningchen: thanks, that's a good point! I just updated it to use `toYaml`

obs-gh-abhishekrao (approved):

PR #351, opened 2026-05-29
init-config: expose all AgentConfig fields as CLI flags + fix fleet default

obs-gh-virjramakrishnan:
## Summary

`init-config` previously exposed CLI flags for ~13 of the ~37 leaf fields in
`AgentConfig`. Anything else (health check, forwarding endpoints, internal
telemetry, exporter batching, etc.) required hand-editing
`/etc/observe-agent/observe-agent.yaml` — making the CLI awkward to drive from
automation (agent skills, runbooks, fleet tooling) and unobvious to users who
don't read the schema.

This PR adds flags for all remaining sections and also fixes a pre-existing
silent-default bug for `--self_monitoring::fleet::enabled`.

### New flags (24)

- **top-level:** `--omit_base_components`, `--attributes` (record-level, distinct from `--resource_attributes`), `--agent_local_file_path`
- **application::RED_metrics:** `--only_generate_for_service_entrypoint_spans`, `--resource_dimensions`, `--span_dimensions`
- **health_check:** `--enabled`, `--endpoint`, `--path`
- **forwarding:** `--enabled`, `--endpoints::http`, `--endpoints::grpc`, `--metrics::convert_cumulative_to_delta`, `--traces::max_span_duration`
- **internal_telemetry:** `--enabled`, `--metrics::{enabled,host,port,level}`, `--logs::{enabled,level,encoding}`
- **host_monitoring::logs::exclude**
- **exporters:** `--sending_queue_batch::{enabled,max_size}`, `--emit_prometheus_target_info_metric`

### Bug fix: `self_monitoring::fleet::enabled` was silently `false` despite cobra default `true`

The flag was registered with cobra default `true` and `BindPFlag`, but lacked a
`v.SetDefault` call. Compare to the adjacent `self_monitoring::enabled` block
which has both. When `init-config` is run without passing the flag,
`viper.Unmarshal` returns the value from `SetViperDefaults` (the Go zero value:
`false`) rather than the cobra default — so the written config has
`fleet.enabled: false` and fleet heartbeats are silently disabled. Discovered
during end-to-end testing where a freshly-installed agent didn't appear in
Fleet Management until we passed `--self_monitoring::fleet::enabled=true`
explicitly.

This PR adds the missing `v.SetDefault("self_monitoring::fleet::enabled", true)`
to match the pattern used for the other Bool flags whose schema lacks a
`default:` tag.

### Refactor

Adds a small `bindFlag(cmd, v, name)` helper that wraps the
`v.BindPFlag(name, cmd.PersistentFlags().Lookup(name))` ritual. Every existing
flag was a two-line dance; the helper keeps the file as a flat list of
declarations and reduces the boilerplate per flag from 2 lines to 1.

## Test plan

- [x] Existing test `Test_InitConfigCommand` updated to reflect the fleet bug fix (cases that didn't explicitly set fleet now expect `Fleet.Enabled=true`).
- [x] Added a new test case to `Test_InitConfigCommand` exercising all new flags whose default is the Go zero value (these play well with `setConfigDefaults` / go-defaults).
- [x] Added `Test_InitConfigCommand_DefaultTrueBoolsCanBeFlippedToFalse` which uses direct YAML regex inspection to verify default-true bools (`health_check::enabled`, `forwarding::enabled`, `internal_telemetry::*::enabled`) actually flip to `false` when the flag is passed. Round-tripping through `AgentConfig` doesn't work for this case because go-defaults treats `false` as a zero value and re-applies the schema's `default:"true"`, hiding the flag-flip from struct-equality assertions.
- [x] `go test ./internal/commands/initconfig/...` passes locally.
- [x] `go build ./...` is clean.

## Followups (not in this PR)

- The struct-tag default mismatch is the root cause of needing per-flag `v.SetDefault` calls in init-config. The cleaner fix is to add `default:"true"` tags to the `Enabled` fields in `configschema.go` (`SelfMonitoringConfig`, `FleetHeartbeatConfig`, `HostMonitoringConfig`, etc.). That would let `defaults.SetDefaults` and viper agree without compensation in the init-config layer. Deliberately not touching the schema in this PR to keep the scope narrow.
- The PR-companion gap analysis ( https://github.com/observeinc/skills/blob/main/.../observe-agent-gaps.md ) notes deeper gaps in `diagnose` (no DNS check, no clock skew check, no fleet-pipeline check, etc.) and missing top-level commands (`config set`, `restart`, `uninstall`). Those are scoped separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/commands/initconfig/initconfig.go
+++ b/internal/commands/initconfig/initconfig.go
@@ -14,23 +14,65 @@ import (
 )
 
 var (
-	config_path                                   string
-	token                                         string
-	observe_url                                   string
-	cloud_resource_detectors                      []string
-	resource_attributes                           map[string]string
+	config_path string
+
+	// Top-level
+	token                 string
+	observe_url           string
+	cloud_resource_detectors []string
+	attributes            map[string]string
+	resource_attributes   map[string]string
+	omit_base_components  bool
+	agent_local_file_path string
+
+	// application.RED_metrics
+	application_RED_metrics_enabled                                       bool
+	application_RED_metrics_only_generate_for_service_entrypoint_spans    bool
+	application_RED_metrics_resource_dimensions                           []string
+	application_RED_metrics_span_dimensions                               []string
+
+	// health_check
+	health_check_enabled  bool
+	health_check_endpoint string
+	health_check_path     string
+
+	// forwarding
+	forwarding_enabled                            bool
+	forwarding_endpoints_http                     string
+	forwarding_endpoints_grpc                     string
 	forwarding_metrics_format                     string
-	application_RED_metrics_enabled               bool
-	self_monitoring_enabled                       bool
+	forwarding_metrics_convert_cumulative_to_delta bool
+	forwarding_traces_max_span_duration           string
+
+	// internal_telemetry
+	internal_telemetry_enabled         bool
+	internal_telemetry_metrics_enabled bool
+	internal_telemetry_metrics_host    string
+	internal_telemetry_metrics_port    int
+	internal_telemetry_metrics_level   string
+	internal_telemetry_logs_enabled    bool
+	internal_telemetry_logs_level      string
+	internal_telemetry_logs_encoding   string
+
+	// self_monitoring
+	self_monitoring_enabled               bool
+	self_monitoring_fleet_enabled         bool
+	self_monitoring_fleet_interval        string
+	self_monitoring_fleet_config_interval string
+
+	// host_monitoring
 	host_monitoring_enabled                       bool
 	host_monitoring_logs_enabled                  bool
 	host_monitoring_logs_include                  []string
+	host_monitoring_logs_exclude                  []string
 	host_monitoring_logs_auto_multiline_detection bool
 	host_monitoring_metrics_host_enabled          bool
 	host_monitoring_metrics_process_enabled       bool
-	self_monitoring_fleet_enabled                 bool
-	self_monitoring_fleet_interval                string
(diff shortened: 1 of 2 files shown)


internal/commands/initconfig/initconfig.go
```diff
@@ -74,64 +116,172 @@ func init() {
...
 
 	cmd.PersistentFlags().StringVar(&observe_url, "observe_url", "", "Observe data collection url")
-	v.BindPFlag("observe_url", cmd.PersistentFlags().Lookup("observe_url"))
+	bindFlag(cmd, v, "observe_url")
 
 	cmd.PersistentFlags().StringSliceVar(&cloud_resource_detectors, "cloud_resource_detectors", []string{}, "The cloud environments from which to detect resources")
-	v.BindPFlag("cloud_resource_detectors", cmd.PersistentFlags().Lookup("cloud_resource_detectors"))
+	bindFlag(cmd, v, "cloud_resource_detectors")
 
-	cmd.PersistentFlags().StringToStringVar(&resource_attributes, "resource_attributes", map[string]string{}, "The cloud environments from which to detect resources")
```
obs-gh-virjramakrishnan, line 93: Seems this description was incorrectly copied? Let me know if the new label makes sense!

obs-gh-mattcotter: Nice catch, lgtm!

obs-gh-mattcotter (approved):
Thanks for cleaning this up!

obs-gh-mattcotter:
Replaced by #358

PR #358, opened 2026-07-20
feat(init-config): use reflection to register flags for all AgentConfig fields

obs-gh-mattcotter:
## Summary

- Replaces the manually-maintained cobra flag list in `init-config` with a recursive reflection walker over `AgentConfig`, so every leaf field automatically gets a flag without manual wiring
- 26 previously-unconfigurable fields now have flags: `health_check`, `internal_telemetry`, `forwarding::endpoints`, `exporters`, `omit_base_components`, full `RED_metrics` sub-config, and more
- A `flagDefaults` map in `initconfig.go` preserves backward-compatible `true` defaults for the four `Enabled` fields that lack a `default` struct tag in the schema (no schema changes)

## Test plan

- [x] Existing `Test_InitConfigCommand` table-driven tests pass unchanged
- [x] New table-driven cases exercise the newly-covered flag paths: forwarding endpoints, internal telemetry, health check endpoint/path, exporters
- [x] `Test_AllConfigFieldsHaveFlags` walks `AgentConfig` via reflection and asserts every handled leaf field has a registered flag — CI will catch any future schema additions that are accidentally left uncovered
- [x] `go build ./...` clean

.... Generated with [Cortex Code](https://docs.snowflake.com/en/user-guide/cortex-code/cortex-code)


internal/commands/initconfig/initconfig.go
```diff
@@ -75,63 +74,127 @@ func init() {
...
 
 func RegisterConfigFlags(cmd *cobra.Command, v *viper.Viper) {
+	// Output-control flags (not part of AgentConfig).
 	cmd.Flags().StringVarP(&config_path, "config_path", "", "", "Path to write config output file to")
 	cmd.Flags().Bool("print", false, "Print the configuration to stdout instead of writing to a file")
 	v.BindPFlag("print", cmd.Flags().Lookup("print"))
 	cmd.Flags().Bool("include-defaults", false, "Include the names and default values for unset config options.")
 	v.BindPFlag("include-defaults", cmd.Flags().Lookup("include-defaults"))
 
-	cmd.PersistentFlags().StringVar(&token, "token", "", "Observe token")
```
obs-gh-virjramakrishnan, line 84: we're losing the help strings here if we delete this: could we keep them somehow?

obs-gh-virjramakrishnan (approved):
LGTM, my agent flagged something small and imo non-blocking, but great otherwise.

PR #366, opened 2026-08-25
feat(config): add confmap converter for legacy component IDs

sfc-gh-kvishwanathan:
The component IDs in our bundled config are the keys that otel_config_overrides and --config merge against, and confmap merges by literal key string. Renaming a bundled ID therefore does not just change a label: an override still naming the old ID stops merging into our component and instead defines a separate one holding only the override's own fields, which then fails validation. This is independent of the upstream type aliases, which keep resolving the old spelling correctly.

Register a confmap.Converter that runs after the resolver has merged every config source and folds legacy IDs back into their canonical counterparts, so existing user configurations survive a rename. This is the same mechanism Splunk (internal/configconverter) and Datadog (comp/otelcol/converter) use.

The mapping table ships empty. It is populated by the PR that performs a rename, which keeps the table correct at every commit.

Upstream provides confmap.Converter explicitly so distributions "build backwards compatible config converters." The resolver merges all URIs, then runs converters (https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/README.md#converter)

Splunk keeps one converter per historical breaking change in internal/configconverter (e.g. loglevel_to_verbosity.go), translating entries in memory without touching user files and warning with the edit to make. They used it to move users from spanmetricsprocessor to spanmetricsconnector, and proposed upstreaming it.

Datadog ships comp/otelcol/converter in DDOT, which also exposes the original and enhanced config. Their need for it drove confmap.Converter support into OCB.

New Relic (NRDOT) renames bundled nodes to track upstream (cumulativetodelta → cumulative_to_delta) and relies on the upstream alias plus release notes — viable only because they hand users a whole config file rather than deep-merging into a bundled one. That is the approach we took, and it is what broke.


internal/configconverter/mappings.go
```diff
@@ -0,0 +1,27 @@
...
+// upstream type alias still resolves it. Keying on the full ID leaves those
+// alone.
+//
+// Add an entry whenever a bundled component ID is renamed, and keep it for at
+// least as long as upstream keeps its own type aliases. Removing an entry is
+// itself a breaking change.
+var LegacyComponentIDs = map[string]string{
+	// Renamed to follow the upstream lower_snake_case component naming
+	// convention (open-telemetry/opentelemetry-collector#14208).
+	"otlphttp/observe":                    "otlp_http/observe",
```
obs-gh-mattcotter, line 21: Can we modify this to split on the `/` and look at the component name? We might want to keep support for `otlphttp` in the observe-agent longer than OTel does upstream.

obs-gh-abhishekrao (commented):
Flagged by a local AI review:

High — Canonical overrides lose precedence
---
[converter.go:112](https://github.com/observeinc/observe-agent/blob/b6b7f405f63ccbcbb51674f284e2385e3f4c004f/internal/configconverter/converter.go#L112)
Bundled templates still use legacy IDs. The converter merges those values over later canonical otel_config_overrides, --config, or --set values. It also emits false deprecation warnings for stock configuration. Rename templates atomically and add a resolver-level precedence test.

E.g.:
Bundled config:
```
exporters:
  otlphttp/observe:
    compression: zstd
```
User override:
```
exporters:
  otlp_http/observe:
    compression: gzip
```
After conversion, the bundled legacy value wins. The effective value remains zstd, silently ignoring the user’s later override.



High — Environment expansion metadata is discarded
---
[converter.go:100](https://github.com/observeinc/observe-agent/blob/b6b7f405f63ccbcbb51674f284e2385e3f4c004f/internal/configconverter/converter.go#L100)
Conf.Get strips ExpandedValue metadata. Numeric or boolean-looking environment variables intended for string fields can consequently fail decoding after remapping. Preserve values through xconfmap.ToStringMapRaw.

E.g.:
```
exporters:
  otlphttp/observe:
    headers:
      x-tenant-id: ${env:TENANT_ID}
```
With `TENANT_ID=123456` Confmap normally remembers that 123456 originated as a string. Conf.Get removes that metadata, leaving an integer. Decoding it into an HTTP header string can then fail during config validate or agent startup.


High — Persistent OTLP queues are stranded
---
[mappings.go:19](https://github.com/observeinc/observe-agent/blob/b6b7f405f63ccbcbb51674f284e2385e3f4c004f/internal/configconverter/mappings.go#L19)
For users configuring sending_queue.storage, changing otlphttp to otlp_http changes the file-storage owner and filename. Existing queued telemetry is no longer drained. The storage ID needs normalization or migration.

E.g.: 
```
exporters:
  otlphttp/observe:
    sending_queue:
      storage: file_storage
```

Suppose an outage leaves 10,000 logs in `exporter_otlphttp_observe_logs`. After upgrading, the canonical OpenTelemetry Protocol (OTLP) exporter ID opens `exporter_otlp_http_observe_logs` and the new queue starts empty. The old logs remain on disk but are never drained.


High — Valid scalar pipeline references are not converted
---
[converter.go:145](https://github.com/observeinc/observe-agent/blob/b6b7f405f63ccbcbb51674f284e2385e3f4c004f/internal/configconverter/converter.go#L145)
OTel accepts scalar/comma-separated references, but this code handles only []any. A scalar legacy reference remains unchanged while its definition is renamed, causing validation failure.

E.g.:
OpenTelemetry accepts this shorthand:
```
service:
  pipelines:
    logs:
      exporters: otlphttp/observe
```
The converter renames the exporter definition but skips this reference because it is a string rather than []any. Startup then fails because otlphttp/observe is referenced but only otlp_http/observe is configured.



Medium — Converter logger settings are ignored
---
[converter.go:44](https://github.com/observeinc/observe-agent/blob/b6b7f405f63ccbcbb51674f284e2385e3f4c004f/internal/configconverter/converter.go#L44)
OTel supplies ConverterSettings.Logger, but the factory discards it and uses the resolution context instead. Startup/reload warnings can bypass the configured collector logger, level, and output.

PR #367, opened 2026-08-26
feat(agent): derive agent instance ID from hostname instead of a rand…

sfc-gh-kvishwanathan:
…om suffix

The agent instance ID was a hostname plus six random characters, so it only stayed the same across restarts because it was written to agent_local_data.json. That made identity depend on a writable data directory, which pods generally do not keep, and left no way for a user or the Helm chart to set the ID.

The ID now resolves from the agent_instance_id config or the OBSERVE_AGENT_INSTANCE_ID environment variable, then any ID already in the local data file so existing installs keep their identity, and finally the hostname. Since every source is stable across restarts, the agent no longer writes the file at all.

### Description

### Checklist


obs-gh-abhishekrao:
Should we consider the `cloud.instance.id` attribute from resourcedetection and use the hostname as a fallback?

internal/agentresource/agentresource.go
```diff
@@ -99,37 +97,38 @@ func (a *AgentResource) GetFilePath() string {
...
 
-func (a *AgentResource) generateAgentInstanceId() string {
-	hostname, err := os.Hostname()
-	if err != nil {
-		hostname = "unknown"
+// configuredInstanceId returns the ID set explicitly by the user, either through
+// the agent config or through the environment variable the collector config and
+// the Helm chart use.
+func configuredInstanceId() string {
+	if id := viper.GetString("agent_instance_id"); id != "" {
```
obs-gh-mattcotter, line 104: I think we set env overrides in viper; is it possible for the viper value to be empty and the env var to be set? Also, instead of accessing this through viper directly we typically use:
```
config.AgentConfigFromViper(viper.GetViper())
```
to handle custom decoding and defaulting.

